PR #10, opened 2020-09-29
fix error condition in ReadOnlyKeywordJsonConverter

shadow-cs:
Wasn't quite sure how you'd test this (I wasn't able to load the `Suite` tests - in VS) so I added just one of the most basic ones I could find. Please let me know how to improve if required.

### DIFF
--- a/JsonSchema/ReadOnlyKeyword.cs
+++ b/JsonSchema/ReadOnlyKeyword.cs
@@ -44,7 +44,7 @@ namespace Json.Schema
 	{
 		public override ReadOnlyKeyword Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
 		{
-			if (reader.TokenType != JsonTokenType.True || reader.TokenType != JsonTokenType.False)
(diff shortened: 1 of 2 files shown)


JsonSchema/ReadOnlyKeyword.cs
```diff
@@ -44,7 +44,7 @@ internal class ReadOnlyKeywordJsonConverter : JsonConverter<ReadOnlyKeyword>
 	{
 		public override ReadOnlyKeyword Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
 		{
-			if (reader.TokenType != JsonTokenType.True || reader.TokenType != JsonTokenType.False)
+			if (reader.TokenType != JsonTokenType.True && reader.TokenType != JsonTokenType.False)
```
shadow-cs, line 47: I'd probably invert the condition altogether, would be a bit easier to read IMO.

gregsdennis:
Thanks for this. I've change the base to a new branch so I can update version and include some other changes.  Just waiting for everything to finish building.

PR #37, opened 2020-11-10
Add Fetch field to ValidationOptions

amosonn:
This makes it more ergonomic to use a custom fetch function.

### DIFF
--- a/JsonSchema.Tests/FetchTests.cs
+++ b/JsonSchema.Tests/FetchTests.cs
@@ -30,6 +30,28 @@ namespace Json.Schema.Tests
 			results.SchemaLocation.Segments.Last().Value.Should().NotBe("$ref");
 		}
 		[Test]
+		public void LocalRegistryFindsRefViaOptions()
+		{
+			var options = new ValidationOptions
+			{
+				OutputFormat = OutputFormat.Detailed,
+                Fetch = uri =>
+                {
+                    if (uri.AbsoluteUri == "http://my.schema/test1")
+                        return JsonSchema.FromText("{\"type\": \"string\"}");
+                    return null;
+                }
+			};
+			var schema = JsonSchema.FromText("{\"$ref\":\"http://my.schema/test1\"}");
+
+			using var json = JsonDocument.Parse("10");
+
+			var results = schema.Validate(json.RootElement, options);
+
+			results.AssertInvalid();
+			results.SchemaLocation.Segments.Last().Value.Should().NotBe("$ref");
+		}
+		[Test]
 		public void GlobalRegistryFindsRef()
 		{
 			var options = new ValidationOptions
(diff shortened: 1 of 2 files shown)


gregsdennis:
I don't think this is useful. The function is defined on the object that is responsible for housing schemas.  Having it also defined here is redundant.

amosonn:
After drafting this, I'm not sure if this is actually the way to go (even though it is done so in some tests). Generally, the user should probably be nudged toward configuring the Global repository rather than the local one, otherwise the fetched Schema will have to re-parsed on each validation. When is a local Registry actually useful?

amosonn:
The idea was though that this lets you define the `Fetch` function directly on construction, instead of accessing it later. You can look at the tests and compare, I think the new format is more elegant.

gregsdennis:
The current architecture caters to multithreaded and repeated validation scenarios better than just a static or singleton registry.  If you want to ensure that only the global registry is fetching schemas, then configure it that way.

I wouldn't be opposed to an option that bypasses the local registry on favor of the global, but I'm not going to completely remove the local registry because one user doesn't see its utility.

amosonn:
Ok, I missed the part where you can keep the `ValidationOptions` around, and then you can have a local registry which is repeatedly used. I would still think that for this use-case, it might be better for `ValidationOptions.SchemaRegistry` to also have a setter, so that you can keep the local registry around even if you want to change the options each time.

Either way, as I said above, the main point of this is to allow setting the `Fetch` function on construction. As long as you don't have a setter for `ValidationOptions.SchemaRegistry`, the registry is essentially attached to the options, so it also makes sense to be able to access it from there. In examples, these are three ways to set the fetch function:

```cs
// 1
var options = new ValidationOptions
{
    Fetch = myFetcher
};

// 2
var options = new ValidationOptions { };
options.SchemaRegistry.Fetch = myFetcher;

// 3
var registry = new SchemaRegistry
{
    Fetch = myFetcher
};
var options = new ValidationOptions
{
    SchemaRegistry = registry
};
```

Option 2 is the only one available now, IMO it's a bit verbose and does not have any advantages over option 1, which is enabled by the PR. Option 3 is more verbose, but IMO cleaner than 2, because the new registry is not implicitly created and can be stored separately from the options; for this a public constructor and a public setter are required.

gregsdennis:
You can already initialize the fetch call when creating the options:

```c#
var options = new ValidationOptions {
    Registry = {
        Fetch = myFetcher
    }
};
```

This works *because* the registry is created upon options construction.  Without that this code doesn't work.

amosonn:
Oh, I wasn't aware of this! I looked only at the code examples in the tests. I guess it would be nice to update the tests to reflect this, but this PR could probably be closed. Sorry for the trouble ;)

PR #62, opened 2021-01-01
Complete json logic

gregsdennis:


gregsdennis:
The analysis revealed that some overloads of `and` and `or` are missing, but the new versions take `params` lists, so there's not breaking change.  Additionally, the nature of this change is such that people likely weren't using the factory functionality, so no harm done.

Issue #79, opened 2021-03-02
Fetch base directory URI from Registry


amosonn:
**Environment**

- Nuget Package: JsonSchema.Net
- Nuget Version: 1.9.0 (regression from 1.8.0)
- OS: Linux
- .Net Target: .net 5.0

**Describe the bug**

In some cases, the schema validation erroneously attempts to fetch the directory containing a schema as a schema itself. It seems to happen when we have one schema referring to another with a relative URI, and the second schema has internal references (i.e. `"#/definitions/...`).

**To Reproduce**

```C#
            var schema1Str = @"
{
  ""$schema"": ""http://json-schema.org/draft-07/schema#"",
  ""$id"": ""schema1.json"",
  ""definitions"": {
    ""myDef"": {
      ""properties"": {
        ""abc"": { ""type"": ""string"" }
      }
    }
  },
  ""$ref"": ""#/definitions/myDef""
}";
            var schema2Str = @"
{
  ""$schema"": ""http://json-schema.org/draft-07/schema#"",
  ""$id"": ""schema2.json"",
  ""$ref"": ""schema1.json""
}";
            var jsonStr = @"{ ""abc"": ""s"" }";
            var schema1 = JsonSerializer.Deserialize<JsonSchema>(schema1Str);
            var schema2 = JsonSerializer.Deserialize<JsonSchema>(schema2Str);
            var json = JsonDocument.Parse(jsonStr).RootElement;
            var uri1 = new Uri("http://first.com/schema1.json");
            var uri2 = new Uri("http://first.com/schema2.json");
            var firstBaseUri = new Uri("http://first.com");
            var map = new Dictionary<Uri, JsonSchema>
            {
                { uri1, schema1 },
                { uri2, schema2 },
            };
            var options = new ValidationOptions
            {
                OutputFormat = OutputFormat.Verbose,
                DefaultBaseUri = firstBaseUri,
                SchemaRegistry =
                {
                    Fetch = uri =>
                    {
                        Console.WriteLine("Fetching {0}", uri);
                        // return map.TryGetValue(uri, out var ret) ? ret : null;
                        return map[uri];
                    }
                }
            };
            var validation = schema2.Validate(json, options);
            Console.WriteLine("Validation: {0}", JsonSerializer.Serialize(validation));
```

Prints
```
Fetching http://first.com/schema1.json
Fetching http://first.com/
[...]

PR #81, opened 2021-03-03
Update GithubTests Issue79

amosonn:
Assert that only the expected uris are fetched.
See #79 #80


gregsdennis:
Will be handled with #82

PR #82, opened 2021-03-04
Fix current uri

amosonn:
This is my attempt to correct the underlying issue in #79 (as mentioned there): sometimes `CurrentUri` was pointing to the parent of the actual current uri.

This is based on my updated test for #79, as in #81, so that one can be merged first or closed and then this one merged directly; and reverts your revert from #80, so that we are using `CurrentUri` in that case, and the test verifies that it is correct.

I'm pretty sure this should also fix diagonstics (verbose results) in some cases, I'll try to provide one later.

PR #83, opened 2021-03-05
Schema 1.9.3

gregsdennis:
Pulls in @amosonn's changes to fix current-uri on the schema context #82.


gregsdennis:
okay.  The analyzer works sufficiently.  I'll have to dig into that later.

PR #84, opened 2021-03-11
remove cmd line args; update config file

gregsdennis:


gregsdennis:
🤔 the `assumeChanges` is a bit noisy flagging all of the libs...

gregsdennis:
I want to be able to run them based on dependencies.  If I change (and update the version for) Pointer, but not Schema, Schema should still be re-analyzed, which would reveal that a bump should occur to account for the bump in Pointer.

So I'd have to encode the lib dependencies into the "run when this path changes" rules.  Even then, it doesn't look like individual jobs can be triggered/skipped based on path (not easily anyway).

PR #98, opened 2021-04-14
Updates to Generation and Schema

### LINKED ISSUE #96: Dynamically modify the childrens properties of JsonSchema
I am currently using your library which work perfectly.

But I have some issues with dynamically modify the childrens properties.

Let me explain, I have two solutions that work, but I can't do everything I want.

1) The first one : 

I use this to generate the skeleton of an object : 

`JsonSchemaBuilder jschemaParent = new JsonSchemaBuilder().FromType<TEntity>();`

This is exactly what I want and work perfectly.

After that I want to be able to modify the childrens properties, and for that the only solution that I found is something like that : 

	IJsonSchemaKeyword proprietes = ((Json.Schema.JsonSchema)jschemaParent).Keywords.First(c => c is PropertiesKeyword);

	Dictionary<string, Json.Schema.JsonSchema> props = new Dictionary<string, Json.Schema.JsonSchema>();

	foreach (var property in ((PropertiesKeyword)proprietes).Properties)
	{
		JsonSchemaBuilder jsonBuilder = new JsonSchemaBuilder();

		foreach (IJsonSchemaKeyword keyword in property.Value.Keywords)
			jsonBuilder.Add(keyword);

		if (property.Key == "Numero")
			jsonBuilder.ReadOnly(true);

		props.Add(property.Key, jsonBuilder);
	}

	jschemaParent.Properties(props);

It works but I have to reimplement each property one by one, by getting each IJsonSchemaKeyword and then adding what I need.

Am I missing something, would there be no way to get a JsonSchemaBuilder from a JsonSchema to do it more easily ?

2) The second one : 

As the first solution I generate the skeleton of an object : 

`var context = SchemaGenerationContextCache.Get(typeof(TEntity), null);`

This also works perfectly. I then created custom Intents to modify children properties :

	PropertiesIntent properties = (PropertiesIntent)context.Intents.First(c => c is PropertiesIntent);

	foreach (var propriete in properties.Properties)
	{
		if (propriete.Key == "Numero")
			propriete.Value.Intents.Add(new ReadOnlyIntent(true));
	}

The problem with this solution is that if the object contains the same type of property several times, it will add the ISchemaKeywordIntent on each property.
For this example, if my object contains two strings, the code above will add two readonly on each strings, but I only want to have it on the property 'Numero' and not the other one.
What am I doing wrong with this solution, how I could avoid duplications of ISchemaKeywordIntent for each property of the same type ?

For any solution I want to generate the skeleton of an object, and then modify children properties, because I have some complex objects, and then modify children properties.

Am I missing something, would there be a simpler solution to do that ?

Thanks in advance and sorry if my english is bad !

### LINKED ISSUE #97: Generating the same scheme FromType multimpe times
**Environment**
- Nuget Package: [JsonSchema.Net, Json.Schema.Generation]
- Nuget Version: JsonSchema.Net -Version 1.10.0, JsonSchema.Net -Version 1.10.0
- OS: Windows 10
- .Net Target: Core 3.1

**Describe the bug**
Generating the same scheme FromType multimpe times and then veryfying some JsonElement with it causes StackOverflow. The place of schema definition doesn't matter: two times in main, one time in main and one time in the callable function, etc.

**To Reproduce**
```
using Json.Schema;
using Json.Schema.Generation;
using System.Text.Json;

public class SimpleValueWidgetSettings
{
    [Required]
    public string name { get; set; }

    [Required]
    public string valueLink { get; set; }

    [Required]
    public string qualityValueLink { get; set; }

    [Required]
    public string unitLink { get; set; }

    [Required]
    public string displayMultiplier { get; set; }

    [Required]
    public string displayUnits { get; set; }

    [Required]
    public string fontColor { get; set; }

    [Required]
    public string stringFormat { get; set; }

    static void Main(string[] args)
    {

        var simpleValueJson = "{\"name\":\"LD2/ATCC1.CtlV.db\"}";
        var simpleValueParsed = JsonDocument.Parse(simpleValueJson).RootElement;
        var simpleValueParsed2 = JsonDocument.Parse(simpleValueJson).RootElement;

        var simpleValueSettingsSchema2 = new JsonSchemaBuilder().FromType<SimpleValueWidgetSettings>().AdditionalProperties(false).Build();
        var simpleValueSettingsSchema = new JsonSchemaBuilder().FromType<SimpleValueWidgetSettings>().AdditionalProperties(false).Build();

        simpleValueSettingsSchema.Validate(simpleValueParsed);
        simpleValueSettingsSchema.Validate(simpleValueParsed2);
    }
}
```
**Expected behavior**
Validating without causing StackOverflow. Multiple definition of the same schema shouldn't cause StackOverflow when trying to validate some JsonElement with the last one.

gregsdennis:
Resolves #96 
Resolves #97


gregsdennis:
Major changes are actually additions of a default param.

PR #110, opened 2021-05-20
Fix #109

### LINKED ISSUE #109: Registration of custom IAttributeHandler ignored
**Environment**
- Nuget Version: 1.3.2
- OS: Windows 10
- .Net Target: .NET Core 3.1, .NET 5.0

**Describe the bug**
I am trying to write an `IAttributeHandler` that will use the [System.ComponentModel.DataAnnotations.DisplayAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.displayattribute) to set a title and description for a schema, but the handler's `AddConstraints` method never seems to get called, despite the handler being registered via the `AttributeHandler.AddHandler<T>` method.

Based on the instructions [here](https://gregsdennis.github.io/json-everything/usage/schema-generation.html#attributes) for writing a handler for an attribute in an external assembly, I can't tell if there is something else that I am supposed to do.

**To Reproduce**
```csharp
void Main() {
    AttributeHandler.AddHandler<DisplayAttributeHandler>();

    var schema = new JsonSchemaBuilder().FromType<MyType>().Build();
   Console.WriteLine(JsonSerializer.Serialize(schema, new JsonSerializerOptions { 
        WriteIndented = true 
    }));

}

public class DisplayAttributeHandler : IAttributeHandler {

    public DisplayAttributeHandler() {
        Console.WriteLine("Constructor called");
    }

    public void AddConstraints(SchemaGeneratorContext context) {
        Console.WriteLine("AddConstraints called");
        // TODO: update the context
    }
    
}

public class MyType {

    [System.ComponentModel.DataAnnotations.Display(Name = "Test Name", Description = "Test description")]
    public string SomeProperty { get; set; }
    
}
```

When I run the above program, I get the following output (note that the `Console.WriteLine` statements in my handler are never called):

```json
{
  "type": "object",
  "properties": {
    "SomeProperty": {
      "type": "string"
    }
  }
}
```

**Expected behavior**
Handlers registered via `AttributeHandler.AddHandler<T>` should be invoked when a schema is being built.

**Additional context**
When I added a `Console.WriteLine` statement to my handler's constructor, I could see that the handler is never even instantiated.

wazzamatazz:
Fixes #109 by ensuring that the check for an existing attribute handler registration is performed correctly.

### DIFF
--- a/JsonSchema.Generation.Tests/AttributeHandlerTests.cs
+++ b/JsonSchema.Generation.Tests/AttributeHandlerTests.cs
@@ -1,0 +1,89 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+using NUnit.Framework;
+
+namespace Json.Schema.Generation.Tests {
+    public class AttributeHandlerTests {
+
+        [Test]
+        public void DirectAttributeHandler() {
+            JsonSchema expected = new JsonSchemaBuilder()
+                .Type(SchemaValueType.Object)
+                .Properties(
+                    ("MyProperty", new JsonSchemaBuilder().Type(SchemaValueType.String).MaxLength(AttributeWithDirectHandler.MaxLength))
+                );
+
+            JsonSchema actual = new JsonSchemaBuilder().FromType<TypeWithCustomAttribute1>();
+
+            Assert.AreEqual(expected, actual);
+        }
+
+
+        [Test]
+        public void IndirectAttributeHandler() {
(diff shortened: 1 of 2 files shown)


gregsdennis (changes requested):
Would you mind adding a simplistic test, please?  Just a new test class that defines a new handler class, registers it, and then generates a schema that would need it.

wazzamatazz:
Tests added 👍

PR #113, opened 2021-05-21
SchemaGen v1.4.0

### LINKED ISSUE #109: Registration of custom IAttributeHandler ignored

### LINKED ISSUE #112: Ordering of properties in auto-generated schemas
**Environment**

When generating a schema using the `FromType<T>` extension for `JsonSchemaBuilder`, the properties are always ordered alphabetically. It would be great to have more control over the order of properties in the resulting schema, or alternatively to add the properties in the order that they have been declared in the `T` type.

For example, if I am building a schema that could be used to automatically generate a form in a UI, I would want the schema properties to be listed in a logical order for the form entry rather than alphabetically. For example:

```csharp
public class MyOptions {
    public string Name { get; set; }
    public string Description { get; set; }
    public bool IsEnabled { get; set; }
    public string Endpoint { get; set; }
    public uint Timeout { get; set; }
}
```

Generated schema:

```json
{
  "type": "object",
  "properties": {
    "Description": {
      "type": "string"
    },
    "Endpoint": {
      "type": "string"
    },
    "IsEnabled": {
      "type": "boolean"
    },
    "Name": {
      "type": "string"
    },
    "Timeout": {
      "type": "integer"
    }
  }
}
```

At the moment, the generated schema orders the properties as `Description`, `Endpoint`, `IsEnabled`, `Name`, `Timeout`, whereas I would ideally want them in the declaration order specified in the class.

gregsdennis:
Depends on #110
Fixes #109
Fixes #112

### DIFF
--- a/JsonSchema.Generation.Tests/AttributeHandlerTests.cs
+++ b/JsonSchema.Generation.Tests/AttributeHandlerTests.cs
@@ -1,0 +1,89 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+using NUnit.Framework;
+
+namespace Json.Schema.Generation.Tests {
+    public class AttributeHandlerTests {
+
+        [Test]
+        public void DirectAttributeHandler() {
+            JsonSchema expected = new JsonSchemaBuilder()
+                .Type(SchemaValueType.Object)
+                .Properties(
+                    ("MyProperty", new JsonSchemaBuilder().Type(SchemaValueType.String).MaxLength(AttributeWithDirectHandler.MaxLength))
+                );
+
+            JsonSchema actual = new JsonSchemaBuilder().FromType<TypeWithCustomAttribute1>();
+
+            Assert.AreEqual(expected, actual);
+        }
+
+
+        [Test]
+        public void IndirectAttributeHandler() {
+            AttributeHandler.RemoveHandler<CustomAttributeHandler>();
+            AttributeHandler.AddHandler<CustomAttributeHandler>();
+
+            JsonSchema expected = new JsonSchemaBuilder()
+                .Type(SchemaValueType.Object)
+                .Properties(
+                    ("MyProperty", new JsonSchemaBuilder().Type(SchemaValueType.String).MaxLength(AttributeWithIndirectHandler.MaxLength))
+                );
+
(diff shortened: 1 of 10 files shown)


gregsdennis:
The version check is wrong.  I have a new type: `PropertyOrder` and a new property that exposes it.  The public API has grown.  That's a minor version. 😕

PR #120, opened 2021-05-25
added uniquekeys vocab draft and metaschema

gregsdennis:
Addresses much of https://github.com/json-schema-org/json-schema-vocabularies/issues/22.


docs_source/usage/vocabs-unique-keys.md
```diff
@@ -0,0 +1,128 @@
...
+
+A draft 2020-12 meta-schema which includes this vocabulary has been defined for convenience.  The ID for the meta-schema is `https://gregsdennis.github.io/json-everything/meta/unique-keys`, and it can also be found at this address.
+
+## 3. The `uniqueKeys` Keyword
+
+### 3.1 Syntax and Semantics
+
+The value of `uniqueKeys` must be an array of JSON Pointers as defined by [RFC 6901](https://tools.ietf.org/html/rfc6901).  At least one JSON Pointer MUST be specified.
+
+These JSON Pointers, when resolved against each item, define a set of values for that item.  If an item does not contain the location specified by a JSON Pointer, that JSON Pointer is skipped and the resulting set does not contain a corresponding value for that JSON Pointer.
```
gregsdennis, line 21: > If an item does not contain the location specified by a JSON Pointer, that JSON Pointer is skipped and the resulting set does not contain a corresponding value for that JSON Pointer.

This may cause problems regarding positioning.  For example, given the pointer set

```
[ "/foo", "/baz", "/bar"]
```

this instance would fail

```json
[
    { "foo": 8, "bar": false },
    { "bar": false, "baz": 8 }
]
```

because both sets would contain `[8, false]` even though the `8` came from `foo` in the first item and `baz` in the second item.  It might make sense to have the implementation recognize a "missing" value, similar to how Javascript uses `undefined`.

docs_source/usage/vocabs-unique-keys.md
```diff
@@ -0,0 +1,128 @@
...
+
+A draft 2020-12 meta-schema which includes this vocabulary has been defined for convenience.  The ID for the meta-schema is `https://gregsdennis.github.io/json-everything/meta/unique-keys`, and it can also be found at this address.
+
+## 3. The `uniqueKeys` Keyword
+
+### 3.1 Syntax and Semantics
+
+The value of `uniqueKeys` must be an array of JSON Pointers as defined by [RFC 6901](https://tools.ietf.org/html/rfc6901).  At least one JSON Pointer MUST be specified.
+
+These JSON Pointers, when resolved against each item, define a set of values for that item.  If an item does not contain the location specified by a JSON Pointer, the implementation MUST keep a placeholder to indicate a missing value.
```
karenetheridge, line 21: By placeholder do you mean a null?

docs_source/usage/vocabs-unique-keys.md
```diff
@@ -0,0 +1,128 @@
...
+
+Validation for this keyword passes if the value sets for all of the items are distinct, that is, no two sets contain the same collection, including missing values.
+
+Implementations SHOULD support JSON Pointers in both the basic and URI formats as specified by RFC 6901.  URI-formatted JSON Pointers are not required in this context, but they are so frequently used throughout JSON Schema that it makes sense for implementations to support them here as well.
```
karenetheridge, line 25: By "URI formatted JSON Pointer" do you mean something like `#/foo/bar`? That's not a JSON pointer anymore though, but a fragment-only uri reference. `#/foo/bar` won't pass the 'json-pointer' format validation, which is what your metaschema uses.

docs_source/usage/vocabs-unique-keys.md
```diff
@@ -0,0 +1,128 @@
...
+
+## 3. The `uniqueKeys` Keyword
+
+### 3.1 Syntax and Semantics
+
+The value of `uniqueKeys` must be an array of JSON Pointers as defined by [RFC 6901](https://tools.ietf.org/html/rfc6901).  At least one JSON Pointer MUST be specified.
+
+These JSON Pointers, when resolved against each item, define a set of values for that item.  If an item does not contain the location specified by a JSON Pointer, the implementation MUST keep a placeholder to indicate a missing value.
+
+Validation for this keyword passes if the value sets for all of the items are distinct, that is, no two sets contain the same collection, including missing values.
```
karenetheridge, line 23: It might be useful to produce an annotation that consists of these value sets -- it could be useful for debugging why something unexpectedly passed validation.

docs_source/usage/vocabs-unique-keys.md
gregsdennis: Actually no.  I mean some internal representation that the value is missing rather than present with `null`.

docs_source/usage/vocabs-unique-keys.md
gregsdennis: May have a misunderstanding here, but I believe that URI formatted pointers are specified in that RFC (section 6, I think), and they follow the same format.  Basically I want to say that both a raw pointer and the format commonly used in JSON Schema should be accepted.

docs_source/usage/vocabs-unique-keys.md
gregsdennis: Do we have other keywords that validate _and_ annotate?

karenetheridge: Most of the applicator keywords?  I don't think it's relevant that they use subschemas as part of their validation -- they still produce a true/false result based on their own unique rules.

docs_source/usage/vocabs-unique-keys.md
karenetheridge: Okay. I guess we can't use null because that could be a legitimate value of a property being referenced.

docs_source/usage/vocabs-unique-keys.md
karenetheridge: Okay, that's what I said/thought.
I think it's weird to accept both, but at any rate, your schema won't validate against that. You'd have to say `"anyOf": [ { "format": "json-pointer"}, {"format":"uri-reference", pattern: "^#/"} ]` (I threw in the pattern for extra clarity).

docs_source/usage/vocabs-unique-keys.md
gregsdennis: I agree with this in principal, but...

1. I'm struggling to find a way for the implementation to represent a missing value.
2. "Unexpectedly passes" would be the only benefit; "unexpectedly fails" wouldn't work because a failure (which I think is the more interesting case) would omit annotations.
3. The annotation would be an array of values _for each item_.  This could get very large very quickly.

PR #123, opened 2021-06-11
Remove development artifact

bastiaantenklooster:


gregsdennis:
Nice catch.  Unfortunately relative JSON pointers don't quite work yet.  It needs some better support from .Net's model first.  They appear to be [working on it](https://github.com/dotnet/runtime/issues/53195) though.

PR #124, opened 2021-06-14
Add support for nullable types

JMPSequeira:
Support Nullable Types in Schema Generation

### DIFF
--- a/JsonSchema.Generation/SchemaGeneratorContext.cs
+++ b/JsonSchema.Generation/SchemaGeneratorContext.cs
@@ -1,4 +1,4 @@
-﻿ using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
@@ -27,6 +27,12 @@ namespace Json.Schema.Generation
 		/// The CLR type currently being processed.
 		/// </summary>
 		public Type Type { get; }
+
+		/// <summary>
+		/// Indicates if the CLR type is originally a Nullable`1.
+		/// </summary>
+		public bool IsNullable { get; }
+
 		/// <summary>
 		/// The set of attributes.  Will be populated when an attribute has a property.
 		/// </summary>
@@ -42,7 +48,9 @@ namespace Json.Schema.Generation
 
 		internal SchemaGeneratorContext(Type type, List<Attribute> attributes, SchemaGeneratorConfiguration configuration)
 		{
-			Type = type;
+			var trueType = Nullable.GetUnderlyingType(type);
+			Type = trueType ?? type;
+			IsNullable = trueType != null;
 			Attributes = attributes;
 			Configuration = configuration;
 		}
@@ -65,8 +73,8 @@ namespace Json.Schema.Generation
 			var thisHash = GetHashCode();
 			var allContexts = GetChildContexts();
 			var defsByHashCode = allContexts.Where(g => g.Value.Count > 1 &&
-			                                            (g.Value.Context.Intents.Count != 1 ||
-			                                            !(g.Value.Context.Intents[0] is TypeIntent)))
+														(g.Value.Context.Intents.Count != 1 ||
+														!(g.Value.Context.Intents[0] is TypeIntent)))
 				.ToDictionary(g => g.Key, g => g.Value.Context);
 
 			var currentNames = new List<string>();
@@ -135,7 +143,7 @@ namespace Json.Schema.Generation
 
 		private Dictionary<int, ContextCount> GetChildContexts()
 		{
-			var contextsToCheck = new List<SchemaGeneratorContext>{this};
+			var contextsToCheck = new List<SchemaGeneratorContext> { this };
 			var contextsReceived = new Dictionary<int, ContextCount>();
 			while (contextsToCheck.Any())
 			{
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,135 @@
+# Rules in this file were initially inferred by Visual Studio IntelliCode from the C:\Users\email\Source\Repos\json-everything codebase based on best match to current usage at 6/14/2021
+# You can modify the rules from these initially generated values to suit your own policies
+# You can learn more about editorconfig here: https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference
+[*.cs]
+
+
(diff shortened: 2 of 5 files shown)


gregsdennis:
This should already be supported through generating `Nullable<T>`.  I'll verify tonight.

JsonSchema.Generation/SchemaGeneratorContext.cs
```diff
@@ -27,6 +27,12 @@ public ContextCount(SchemaGeneratorContext context)
 		/// The CLR type currently being processed.
 		/// </summary>
 		public Type Type { get; }
+
+		/// <summary>
+		/// Indicates if the CLR type is originally a Nullable`1.
+		/// </summary>
+		public bool IsNullable { get; }
```
gregsdennis, line 34: I don't like the idea of putting this on the context. It's a separate type.  I'd like to see how it behaves currently.

Furthermore, a nullable `int` should render as `"type": ["integer", "null"]`.  And if that's supported, should reference types have the same multi-type support? Maybe there needs to be an option that enables this kind of support.

JMPSequeira:
> This should already be supported through generating `Nullable<T>`. I'll verify tonight.

Got carried away. Nice lib btw.
I intend to use it on a personal project and noticed that nullable types where being generated as objects.

gregsdennis (commented):
How was the editor config generated? Does it match with my R# rules?

JMPSequeira:
> How was the editor config generated? Does it match with my R# rules?

Inferred from code analysis by VS

JsonSchema.Generation/SchemaGeneratorContext.cs
JMPSequeira: > I don't like the idea of putting this on the context. It's a separate type. I'd like to see how it behaves currently.

It was the minimal disruption route.

> Furthermore, a nullable `int` should render as `"type": ["integer", "null"]`. 

Agreed. And shouldn't an `int` be marked required by default?

> And if that's supported, should reference types have the same multi-type support? Maybe there needs to be an option that enables this kind of support.

Reference types should render as `"type":[{referenceType}, "null"]` unless stated as Required, or, if Nullable reference types are enabled, not marked as nullable.

gregsdennis: > And shouldn't an `int` be marked required by default?

I want explicitly required props via `[Required]`.  Deserializing something that's missing a given prop would just result in the default value, and this is usually fine.

> Reference types should render as `"type":[{referenceType}, "null"]`

Here's where .Net's and JSON Schema's concepts of "required" interfere.  .Net's interprets absent and null (or default, rather) as the same.  JSON Schema separates these ideas so that I can require a value to be present but optionally specify that it could be null.  Example:

```c#
[Required]
public string Foo { get; set; }
```

could generate

```json
{
  "properties": {
    "foo": { "type": [ "string", "null" ] }
  },
  "required": [ "foo" ]
}
```

but this schema would allow `{ "foo": null }` which would fail .Net's validation.

I can see the logic behind having _either_ `required` or including `type: null` from the C# perspective, but that leaves some open questions.  For this table, I've listed what _seems_ to be the the proper behavior, but some (❓s) I'm unsure on.

| | value | value? | reference |
|---|---|---|---|
| required | exclude `null` ✔️<br>list in `required` ✔️ | include `null` ❓<br>list in `required` ✔️ | exclude `null` ❓<br>list in `required` ✔️ |
| optional | exclude `null` ✔️<br>exclude from `required` ✔️ | include `null` ✔️<br>exclude from `required` ✔️ | include `null` ✔️<br>exclude from `required` ✔️ |

> ... if Nullable reference types are enabled, not marked as nullable.

I can't tell from this library if the consumer has nullable reference types enabled.  I can check for the attribute, but if it's not there, I can't make assumptions because either they are using NRT and it's not nullable, or they're not using NRT.

gregsdennis:
I think I'd like to see what an implementation of a generator that covers `Nullable<T>` would look.  This is the approach I'd default to.

JsonSchema.Generation/SchemaGeneratorContext.cs
gregsdennis: Regarding that last bit, there's a lot of chatter on recognizing the attribute for when NRT are enabled, but nothing that says what happens in C#8 if  NRT are disabled.  I'll have to run an experiment.

https://stackoverflow.com/questions/58453972/how-to-use-net-reflection-to-check-for-nullable-reference-type
https://github.com/RicoSuter/Namotion.Reflection
https://github.com/dotnet/docs/issues/9662
https://codeblog.jonskeet.uk/2019/02/10/nullableattribute-and-c-8/

JMPSequeira: It still captures the null intent, 

From what I could gather from [NJsonSchema](https://github.com/RicoSuter/NJsonSchema/wiki/JsonSchemaGenerator) and [Newtonsoft](https://www.newtonsoft.com/jsonschema/help/html/GeneratingSchemas.htm) the null handling is configurable.

IMHO it should be about capturing user intent, from a C#(going out on a limb here assuming most generation is to provide clients with validation logic) and zero-config perspective, making it as close to standard validation in .NET. This does not mean it can't be configurable, just that I'd make the defaults reflect this line of thought.

Either way, having it configurable would solve your doubts on how to handle it.

gregsdennis: > making it as close to standard validation in .NET

This is where my philosophy departs from these libraries.  I want to capture JSON Schema as purely as possible.  I'm not so concerned about how standard validation works.  My concern is how JSON Schema wants validation to work.

The purpose of this generation library is to create a JSON Schema starting point that can be further refined by humans.  It's definitely **not** intended to be used to generate production-ready schemas.  As such, I'd like to see this entire idea covered under a configuration option.

I think it's a good idea, though.  I'd just like to take a different approach to solve it.

JMPSequeira:
Implemented with a nullable Generator here. Check it out:
https://github.com/gregsdennis/json-everything/compare/master...JMPSequeira:Support_Nullable_Value_Types?expand=1

gregsdennis:
This has been good stepping stone to #127 and leading on to #128.  Thanks for your work on this.  Closing this PR for now.

PR #125, opened 2021-06-14
Only generate properties if properties exist

JMPSequeira:
Classes with no properties, or objects generate an empty properties object.
PR to fix that


gregsdennis:
Some of the actions don't work on fork PRs. ???

gregsdennis (approved):

PR #128, opened 2021-06-18
Nullable support using a refiner

gregsdennis:
A third attempt at supporting nullable types.


JMPSequeira:
Nice work.

I'm really glad this went through 😃

PR #132, opened 2021-06-19
Fix memory leaks.

ddunkin:
JsonDocument rents from ArrayPool and must be disposed to prevent memory leaks.

### DIFF
--- a/Json.More/JsonElementExtensions.cs
+++ b/Json.More/JsonElementExtensions.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Text.Json;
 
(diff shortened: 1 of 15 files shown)


Json.More/JsonElementExtensions.cs
```diff
@@ -181,8 +182,8 @@ public static JsonElement AsJsonElement(this short value)
 		/// <remarks>This is a workaround for lack of native support in the System.Text.Json namespace.</remarks>
 		public static JsonElement AsJsonElement(this bool value)
 		{
-			var doc = JsonDocument.Parse($"{value.ToString().ToLower()}");
-			return doc.RootElement;
+			using var doc = JsonDocument.Parse(value ? "true" : "false");
```
ddunkin, line 185: This could further be optimized by returning the same static true or false instances of `JsonElement`.

Json.More/JsonElementExtensions.cs
```diff
@@ -205,8 +206,8 @@ public static JsonElement AsJsonElement(this decimal value)
 		/// <remarks>This is a workaround for lack of native support in the System.Text.Json namespace.</remarks>
 		public static JsonElement AsJsonElement(this double value)
 		{
-			var doc = JsonDocument.Parse($"{value}");
-			return doc.RootElement;
+			using var doc = JsonDocument.Parse(value.ToString(CultureInfo.InvariantCulture));
```
ddunkin, line 209: `value.ToString(CultureInfo.InvariantCulture)` prevents unexpected encoding in different locales, e.g. `,` instead of `.`.
https://replit.com/@ddunkin/NET-doubleToString

gregsdennis:
This looks good.  Thanks for the improvements.

PR #135, opened 2021-06-29
Schemagen camel case

### LINKED ISSUE #134: JsonSchema properties in camelcase
Hi, 

In my solution I use the CamelCase format for our response.
For example, this object  :

```c#
public class Client
{
    public int Id { get; set; }
    public string Name { get; set; }
    public string OtherName { get; set; }
}
```

will be shown as 

```json
{
    "id": 123,
    "name": "name",
    "otherName": "name2"
}
```

in my api.

In addition I use your library to apply some JsonSchema, but the problem is that the properties are in PascalCase.
```json
{
  "type": "object",
  "properties": {
        "Id": {
          "type": "integer"
        },
        "Name": {
          "type": "string"
        },
		"OtherName": {
          "type": "string"
        }
  },
  "required": [
    "Id",
    "Name"
  ]
}
```

I was supposing that using 

```c#
var serializerOptions = new JsonSerializerOptions { WriteIndented = true, PropertyNamingPolicy = JsonNamingPolicy.CamelCase };
```

will do the job, but it actually doesn't work as expected.

Is there a solution to have in my JsonSchema something like this : 

```json
{
  "type": "object",
  "properties": {
        "id": {
          "type": "integer"
        },
        "name": {
          "type": "string"
        },
		"otherName": {
          "type": "string"
        }
  },
  "required": [
    "Id",
    "Name"
  ]
}
```

gregsdennis:
Resolves #134

### DIFF
--- a/JsonSchema.Generation.Tests/PropertyNamingTests.cs
+++ b/JsonSchema.Generation.Tests/PropertyNamingTests.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using NUnit.Framework;
+
+using static Json.Schema.Generation.Tests.AssertionExtensions;
+
+namespace Json.Schema.Generation.Tests
+{
+	public class PropertyNamingTests
+	{
+		class Target
+		{
+			public string PropertyThatNeeds_Changing { get; set; }
+		}
+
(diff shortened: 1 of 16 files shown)


gregsdennis:
Dangit... the "assume changes" argument makes these post even if there are no changes.

PR #136, opened 2021-06-30
Order properties by type and then declaration order

wazzamatazz:
See [this comment](https://github.com/gregsdennis/json-everything/issues/112#issuecomment-845935019) in #112 for more details.

This PR extends the "as declared" property ordering to ensure that members are ordered from least-derived type to most-derived type, and then in declaration order in each individual type.

Consider the following types:

```csharp
public class BaseClass {
  public int First { get; set; }
  public int Second { get; set; }
}

public class DerivedClass : BaseClass {
  public int Third { get; set; }
  public int Fourth { get; set; }
}
```

When generating a schema for `DerivedClass` using `PropertyOrder.AsDeclared`, the order of properties in the schema is currently:

- `Third`
- `Fourth`
- `First`
- `Second`

This PR ensures that the properties from `BaseClass` are listed before `DerivedClass` i.e.

- `First`
- `Second`
- `Third`
- `Fourth`

The issue linked above mentions that a refiner could be used to accomplish the same task, but a general-purpose solution for any given type would effectively need to follow the same logic as the PR adds, but without the benefit of access to the `MemberInfo.MetadataToken` that `ObjectSchemaGenerator` already uses to determine the order at present, so would be more efficient to do this in `ObjectSchemaGenerator` instead.

Currently, a new comparer object is created for each call to `AddConstraints` using the `SchemaGeneratorContext.Type` property, but a cache of instantiated comparers could be stored in e.g. a `ConcurrentDictionary` is preferred for performance reasons.

### DIFF
--- a/JsonSchema.Generation/Generators/ObjectSchemaGenerator.cs
+++ b/JsonSchema.Generation/Generators/ObjectSchemaGenerator.cs
@@ -34,7 +34,7 @@ namespace Json.Schema.Generation.Generators
 
 			membersToGenerate = context.Configuration.PropertyOrder switch
 			{
-				PropertyOrder.AsDeclared => membersToGenerate.OrderBy(m => m, context.AsDeclaredMemberInfoComparer.Value),
+				PropertyOrder.AsDeclared => membersToGenerate.OrderBy(m => m, context.DeclarationOrderComparer.Value),
 				PropertyOrder.ByName => membersToGenerate.OrderBy(m => m.Name),
 				_ => membersToGenerate
 			};
--- a/JsonSchema.Generation/Generators/ObjectSchemaGenerator.cs
+++ b/JsonSchema.Generation/Generators/ObjectSchemaGenerator.cs
@@ -97,16 +97,19 @@ namespace Json.Schema.Generation.Generators
 
 			private static bool HasMetadataToken(MemberInfo? member)
 			{
-				if (member == null)
-				{
-					return false;
-				}
+				if (member == null) return false;
 
 #if NET5_0_OR_GREATER
 				return member.HasMetadataToken();
 #else
-				try { var token = member.MetadataToken; return true; }
-				catch (InvalidOperationException) { return false; }
+				try 
+				{ 
+					var token = member.MetadataToken; return true; 
+				}
+				catch (InvalidOperationException) 
+				{ 
+					return false; 
+				}
 #endif
 			}
 
@@ -117,20 +120,11 @@ namespace Json.Schema.Generation.Generators
 
 			public override int Compare(MemberInfo? x, MemberInfo? y)
 			{
-				if (x == y)
-				{
-					return 0;
-				}
+				if (x == y) return 0;
+				
+				if (x == null) return 1;
 
-				if (x == null)
-				{
-					return 1;
-				}
-
-				if (y == null)
-				{
-					return -1;
-				}
+				if (y == null) return -1;
 
 				// Get metadata tokens for the types that declared the members.
 				var xTypeToken = GetMetadataToken(x.DeclaringType);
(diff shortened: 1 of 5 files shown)


gregsdennis:
Thanks for this. I'd like to go over this in finer detail, but just to start...

> but a cache of instantiated comparers could be stored...

Instead of doing this separately, we could just add the converter to the context as an internally accessible property (lazy loaded).  It's already cached and (probably sufficiently) unique per type.

---

The build failures are due to this being a PR from a fork.

JsonSchema.Generation/Generators/ObjectSchemaGenerator.cs
```diff
@@ -71,9 +71,106 @@ public void AddConstraints(SchemaGeneratorContext context)
...
+			{
+				return new MemberInfoMetadataTokenComparer(type ?? throw new ArgumentNullException(nameof(type)));
+			}
+
+			private static bool HasMetadataToken(MemberInfo? member)
+			{
+				if (member == null)
+				{
+					return false;
+				}
```
gregsdennis, line 103: Please remove braces and place on a single line.  (Other places too.)

JsonSchema.Generation/Generators/ObjectSchemaGenerator.cs
```diff
@@ -71,9 +71,106 @@ public void AddConstraints(SchemaGeneratorContext context)
...
+
+#if NET5_0_OR_GREATER
+				return member.HasMetadataToken();
+#else
+				try { var token = member.MetadataToken; return true; }
+				catch (InvalidOperationException) { return false; }
```
gregsdennis, line 109: Conversely, please break this into multiple lines, if-else style.

gregsdennis (changes requested):

wazzamatazz:
> Instead of doing this separately, we could just add the converter to the context as an internally accessible property (lazy loaded). It's already cached and (probably sufficiently) unique per type.

Have added a commit to move the comparer class into `SchemaGeneratorContext` and add a lazy-loaded internal property that will create the comparer for the context's type if/when required.

wazzamatazz:
Would you prefer if the comparer was an internal class in its own file? I placed it near the top of `SchemaGeneratorContext` (next to the other private class declaration) but it means that you have to do a bit of scrolling before you get to the actual `SchemaGeneratorContext` code itself.

gregsdennis:
> Would you prefer if the comparer was an internal class in its own file?

That's what I had in mind, yes.

JsonSchema.Generation/SchemaGeneratorContext.cs
```diff
@@ -24,6 +25,91 @@ public ContextCount(SchemaGeneratorContext context)
...
+					type = type.BaseType!;
+				} while (type != null);
+
+				_typeOrder = typeStack.Select(GetMetadataToken).ToArray();
+			}
+
+			public static MemberInfoMetadataTokenComparer ForType(Type type)
+			{
+				return new MemberInfoMetadataTokenComparer(type ?? throw new ArgumentNullException(nameof(type)));
+			}
```
gregsdennis, line 49: For internal things, I think it's okay to just expose the constructor.  I don't see much benefit in this pattern when it just creates a new one anyway.

JsonSchema.Generation/SchemaGeneratorContext.cs
```diff
@@ -40,12 +126,17 @@ public ContextCount(SchemaGeneratorContext context)
 		/// The generator configuration.
 		/// </summary>
 		public SchemaGeneratorConfiguration Configuration { get; }
+		/// <summary>
+		/// <see cref="IComparer{MemberInfo}"/> for ordering members when generating a schema with <see cref="PropertyOrder.AsDeclared"/> ordering enabled.
+		/// </summary>
+		internal Lazy<IComparer<MemberInfo>> AsDeclaredMemberInfoComparer { get; }
```
gregsdennis, line 132: A [poor-man's lazy](https://stackoverflow.com/a/63085705/878701) here instead (and a simpler name), please.  This doesn't expose the lazy nature, and we can just use the value we need.  There's also a bit of overhead for `Lazy<T>` that we can avoid.  I'm not worried about thread safety here since this object should never see multiple threads.

```suggestion
		private IComparer<MemberInfo> _memberInfoComparer;
		internal IComparer<MemberInfo> AsDeclaredMemberInfoComparer => _memberInfoComparer ??= new MemberInfoMetadataTokenComparer(Type);
```

wazzamatazz: Internal property name is now `DeclarationOrderComparer`

gregsdennis:
I'm going to pull this into a new branch and make a few edits including updating the version and release notes.  Then this should be ready to go.

Thanks for the help.

wazzamatazz:
Great, thanks!

gregsdennis:
It's already published and ready for you.

PR #139, opened 2021-07-02
Fix #138

### LINKED ISSUE #138: Type initialiser for SchemaRegistry throws an exception
**Environment**
- Nuget Version: 1.10.5
- OS: Windows 10
- .Net Target: .NET Core 3.1.16, .NET 5.0.7

**Describe the bug**
An exception is thrown by the type initialiser for `ValidationOptions` that is caused by a `NullReferenceException` in the `SchemaRegistry` type initialiser:

```
System.TypeInitializationException: The type initializer for 'Json.Schema.ValidationOptions' threw an exception.
 ---> System.TypeInitializationException: The type initializer for 'Json.Schema.SchemaRegistry' threw an exception.
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Json.Schema.JsonSchema.RegisterSubschemasAndGetBaseUri(SchemaRegistry registry, Uri currentUri)
   at Json.Schema.ItemsKeyword.Json.Schema.IRefResolvable.RegisterSubschemas(SchemaRegistry registry, Uri currentUri)
   at Json.Schema.JsonSchema.RegisterSubschemasAndGetBaseUri(SchemaRegistry registry, Uri currentUri)
   at Json.Schema.DefinitionsKeyword.Json.Schema.IRefResolvable.RegisterSubschemas(SchemaRegistry registry, Uri currentUri)
   at Json.Schema.JsonSchema.RegisterSubschemasAndGetBaseUri(SchemaRegistry registry, Uri currentUri)
   at Json.Schema.JsonSchema.RegisterSubschemas(SchemaRegistry registry, Uri currentUri)
   at Json.Schema.SchemaRegistry..cctor()
   --- End of inner exception stack trace ---
   at Json.Schema.SchemaRegistry..ctor(ValidationOptions options)
   at Json.Schema.ValidationOptions..ctor()
   at Json.Schema.ValidationOptions..cctor()
   --- End of inner exception stack trace ---
   at Json.Schema.ValidationOptions.get_Default()
```

The issue means that it is not possible to validate a `JsonElement` against a schema.

**To Reproduce**
```csharp
try {
    var options = Json.Schema.ValidationOptions.Default;
}
catch (Exception e) {
    Console.WriteLine(e.ToString());
}
```

**Expected behavior**
The type initialiser for `SchemaRegistry` should not throw a `NullReferenceException`.

wazzamatazz:
Fixes #138 by making `SchemaRegistry.ValidatingAs` get its value from `_options` every time and protecting against `_options` being `null`.

### DIFF
--- a/JsonSchema/SchemaRegistry.cs
+++ b/JsonSchema/SchemaRegistry.cs
@@ -49,7 +49,10 @@ namespace Json.Schema
 			set => _fetch = value;
 		}
 
-		internal Draft ValidatingAs => _options.ValidatingAs;
+		internal Draft ValidatingAs
+		{
+			get => _options?.ValidatingAs ?? Draft.Unspecified;
+		}
 
 		static SchemaRegistry()
 		{


JsonSchema/SchemaRegistry.cs
```diff
@@ -49,7 +49,10 @@ private class Registration
 			set => _fetch = value;
 		}
 
-		internal Draft ValidatingAs => _options.ValidatingAs;
+		internal Draft ValidatingAs
+		{
+			get => _options?.ValidatingAs ?? Draft.Unspecified;
```
gregsdennis, line 54: I'm more interested in giving why a registration can have a null options.  That shouldn't happen.

gregsdennis (changes requested):

JsonSchema/SchemaRegistry.cs
wazzamatazz: I've explained this in #138 [here](https://github.com/gregsdennis/json-everything/issues/138#issuecomment-872855543). It's because of how the property is declared.

The following two property declarations are equivalent:

```csharp
internal Draft ValidatingAs => _options.ValidatingAs;

internal Draft ValidatingAs { get; } = _options.ValidatingAs;
```

Since this is a property with only a getter and an initialiser, the initialiser runs when the object is created, but before the constructor is called. The `_options` field does not get set until the constructor is called, meaning that it is still `null` when the property initialiser code runs.

wazzamatazz: I think that, technically, the change I have made makes the null check redundant, because the property value is no longer being set before the constructor is called, meaning that `_options` will always be non-null by the time that the getter is called.

wazzamatazz: Yeah I think I was too quick to blame the property initialiser syntax. Will let you dig into how/why `_options` can be `null`.

wazzamatazz:
Closing this - the PR does not address the root cause of the problem.

PR #142, opened 2021-07-08
added steps to check for changes

gregsdennis:
this shouldn't post analysis reports


gregsdennis:
okay... it should only post for 'more' project

Issue #23, opened 2020-10-09
"Diff"-ing and other Patch extensions


gregsdennis:
A library that can compare two JSON documents and report on their differences.

gregsdennis:
I think this may actually be a good extension to the Patch library.  It could analyze the differences and generate JSON Patch documents that represent the transformation from one to the other.

gregsdennis:
Two more extensions:

- Inversions (undo)
- Condensing (merging two patches)

LordXaosa:
Hi!
I've got a solution how can I contribute?

weichch:
> I think this may actually be a good extension to the Patch library. It could analyze the differences and generate JSON Patch documents that represent the transformation from one to the other.

I'm actually thinking the same but the other way around. 😃 

I have a [diff library](https://github.com/weichch/system-text-json-jsondiffpatch) that can produce JSON Patch document in RFC format but patching using such format is not yet support. This brought me here for alternatives.

gregsdennis:
Someone contributed some code a while ago that can create a patch from two documents (and even two objects), but there was no intermediate diff, so I left this open.

JacobAtchley:
seems like this is a common thing the community is trying to wrangle 

we also made our own here 

https://github.com/firebend/json-patch-generator

PR #163, opened 2021-09-13
Patch diff

gregsdennis:
Related to #23 - Create patches from two JSON instances / objects.


JsonPatch/PatchExtensions.cs
```diff
@@ -7,136 +7,157 @@
...
+		/// </summary>
+		/// <param name="patch">The patch</param>
+		/// <param name="obj">The object</param>
+		/// <typeparam name="TOriginal">The type of the original object.</typeparam>
+		/// <typeparam name="TTarget">The type of the target object.</typeparam>
+		/// <returns>New instance of patched object</returns>
+		/// <exception cref="InvalidOperationException">Thrown when the patch cannot be applied.</exception>
+		public static TTarget Apply<TOriginal, TTarget>(this JsonPatch patch, TOriginal obj)
+		{
+			using var doc = JsonDocument.Parse(JsonSerializer.Serialize(obj));
```
LordXaosa, line 39: Could you please add JsonSerializer options here? I think it's necessary in all serialization\deserialization code.

JsonPatch/PatchExtensions.cs
```diff
@@ -7,136 +7,157 @@
...
-                if (i >= mod.GetArrayLength())
-                {
-                    patch.Add(PatchOperation.Remove(path.Combine(i)));
-                    continue;
-                }
+		private static void PatchForArray(JsonElement orig, JsonElement mod, List<PatchOperation> patch, JsonPointer path)
+		{
+			for (int i = 0; i < Math.Max(orig.GetArrayLength(), mod.GetArrayLength()); i++)
+			{
+				var ui = (uint) i;
```
LordXaosa, line 133: Why not implicit int conversion? Here you have to make 2 casts. From int to uint, and then to string inside. Why not directly into to string? Just thoughts

JsonPatch/PatchExtensions.cs
gregsdennis: This is putting submitting an object and returning an object.  The intermediate patch should make no difference.

JsonPatch/PatchExtensions.cs
gregsdennis: It's not proper for JSON Pointer to accept a negative number.

[RFC 6901 section 4](https://datatracker.ietf.org/doc/html/rfc6901#section-4)

> If the currently referenced value is a JSON array, the reference token MUST contain either:
> 
> - characters comprised of digits (see ABNF below; note that leading zeros are not allowed) that represent an **_unsigned base-10 integer value_**, making the new referenced value the array element with the zero-based index identified by the token, or
> - exactly the single character "-", making the new referenced value the (nonexistent) member after the last array element.

JsonPatch/PatchExtensions.cs
LordXaosa: Sometimes there could be situation where object need special converter and it could be passed over options. Without it serialization could be impossible. For example in case of complex objects that are used in external library and there no defined converter for such model, but it could be used.
Also I mentions earlier, JsonPatch could be from web API which would be in camelCase, but default serialization would be in PascalCase and won't working.

JsonPatch/PatchExtensions.cs
LordXaosa: Thanks for clarification. It's fair enough

JsonPatch/PatchExtensions.cs
```diff
@@ -0,0 +1,164 @@
...
+		/// <typeparam name="TTarget">The type of the target object.</typeparam>
+		/// <returns>New instance of patched object</returns>
+		/// <exception cref="InvalidOperationException">Thrown when the patch cannot be applied.</exception>
+		public static TTarget Apply<TOriginal, TTarget>(this JsonPatch patch, TOriginal obj, JsonSerializerOptions? options = null)
+		{
+			using var doc = JsonDocument.Parse(JsonSerializer.Serialize(obj, options));
+			var patchResult = patch.Apply(doc.RootElement);
+			if (!patchResult.IsSuccess)
+				throw new InvalidOperationException($"{patchResult.Error} Operation: {patchResult.Operation}");
+			var result = JsonSerializer.Deserialize<TTarget>(patchResult.Result.GetRawText());
```
LordXaosa, line 44: Deserialization need options too, just in case, I guess.

gregsdennis: Good catch.  I hope one options will suffice for both operations...

LordXaosa: I hope so too, but if it's optional may be splitting could be more convenient?

gregsdennis: I'm not fussed about it... most options are useful for one or the other direction, but not both.

LordXaosa: I'm just assuming that we could got Source from one library and Target to other library object, so we may need different converters... I guess it would be rare situation but if would be - single options could be not such goot thing...
I think you could use something like this
`var result = JsonSerializer.Deserialize<TTarget>(patchResult.Result.GetRawText(), deserializeOptions??options);`
Just suggesting, not infusing :)

gregsdennis:
You should be able to test this at https://json-everything.net/json-patch in about 10 minutes.

gregsdennis:
Oh.... I guess not.  I need to build UI for this new functionality.

PR #175, opened 2021-11-04
Make initialization of the static classes thread-safe

ranma42:
While using your library in an asp.net core project I hit some unexpected failures. The only reproducer I managed to find out required concurrent validations, which pointed in the direction of some issue in the thread-safety of the library.

I am quite confident that the issue stems from

https://github.com/gregsdennis/json-everything/blob/f52854937a06afa87794345cb8e66bb1d2d45ee8/JsonSchema/SchemaRegistry.cs#L66-L73

as this makes it possible for two threads A & B to perform the following:
 - A proceeds up to line 72, yields;
 - B finds `Global._options != null` at line 66, `return`s and performs the validation (without the MetaSchemas registrations)
 - A finally continues with line 73 and beyond

A fix for this race condition could be implemented by moving line `71` to the very bottom of the `lock` block, but I believe that a (minor) refactoring of the code allows for a cleaner initialization of both `SchemaRegistry` and `ValidationOptions`.

The approach followed in this branch involves:
 - removing all of the static initialization/construction from `SchemaRegistry`
 - adding a static constructor to `ValidationOptions` that performs the metaschema initialization on the registry associated to the default validation options (aka the global registry)
 - making the global registry available through a getter

As a (minor) nice side effect, this branch makes it possible to flag `SchemaRegistry._options` as `readonly` and make it non-nullable :)

### DIFF
--- a/JsonSchema/JsonSchema.cs
+++ b/JsonSchema/JsonSchema.cs
@@ -98,7 +98,6 @@ namespace Json.Schema
 		public ValidationResults Validate(JsonElement root, ValidationOptions? options = null)
 		{
 			options = ValidationOptions.From(options ?? ValidationOptions.Default);
-			SchemaRegistry.Initialize();
 
 			var context = new ValidationContext(options)
 				{
--- a/JsonSchema/SchemaRegistry.cs
+++ b/JsonSchema/SchemaRegistry.cs
@@ -32,7 +32,7 @@ namespace Json.Schema
 		private Dictionary<Uri, Registration>? _registered;
 		private Func<Uri, JsonSchema?>? _fetch;
 		private Stack<Uri>? _scopes;
-		private ValidationOptions? _options;
+		private readonly ValidationOptions _options;
 
 		/// <summary>
 		/// The global registry.
@@ -48,7 +48,7 @@ namespace Json.Schema
 			set => _fetch = value;
 		}
 
-		internal Draft ValidatingAs => _options!.ValidatingAs;
+		internal Draft ValidatingAs => _options.ValidatingAs;
 
 		internal SchemaRegistry(ValidationOptions options)
 		{
--- a/JsonSchema/SchemaRegistry.cs
+++ b/JsonSchema/SchemaRegistry.cs
@@ -37,7 +37,7 @@ namespace Json.Schema
 		/// <summary>
 		/// The global registry.
 		/// </summary>
-		public static SchemaRegistry Global { get; }
+		public static SchemaRegistry Global => ValidationOptions.Default.SchemaRegistry;
 
 		/// <summary>
 		/// Gets or sets a method to enable automatic download of schemas by `$id` URI.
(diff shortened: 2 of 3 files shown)


gregsdennis:
Thanks for looking into this.  I've had some difficulties with the timing in here before.

Please take a look at the [history of `SchemaRegistry`](https://github.com/gregsdennis/json-everything/commit/701891a9403cad34def9caffc131b81fef06106b#diff-cd6efe3d90e98f5d95943bea9c4f82def30fcf2d5ed36f5e1b1607df9ae978d6).  You can see that it used to be pretty close to what you're proposing.

ranma42:
> Thanks for looking into this. I've had some difficulties with the timing in here before.
> 
> Please take a look at the [history of `SchemaRegistry`](https://github.com/gregsdennis/json-everything/commit/701891a9403cad34def9caffc131b81fef06106b#diff-cd6efe3d90e98f5d95943bea9c4f82def30fcf2d5ed36f5e1b1607df9ae978d6). You can see that it used to be pretty close to what you're proposing. There's a lot of discussion in #138 as well.

I have looked a bit into the [history](https://github.com/gregsdennis/json-everything/commits/master/JsonSchema/SchemaRegistry.cs), but it looks like all of the attempts were problematic, mostly because of a cyclic dependency between `SchemaRegistry` and `ValidationOptions`:
 - https://github.com/gregsdennis/json-everything/blob/29c736ba99db4117b5dc6ed1fdaea08096655878/JsonSchema/SchemaRegistry.cs is the current version, which avoids the cyclic dependency at the cost of a very convoluted construction which results in the race condition described above;
 - https://github.com/gregsdennis/json-everything/blob/29c736ba99db4117b5dc6ed1fdaea08096655878/JsonSchema/SchemaRegistry.cs was hacking around the issue by making `_options` mutable and setting it upon access to `Global`;
 - I believe that when you say "that it used to be pretty close to what you're proposing" you are referring to https://github.com/gregsdennis/json-everything/blob/068f158830d4aa2e61862ebe10fc161e69d1f546/JsonSchema/SchemaRegistry.cs as it tries to perform the initialization with no trick.

I believe that all of these implementation share an issue that is not present after my patch, i.e. all of them try to "do stuff" in the static initializer/constructor of `SchemaRegistry`.
Although the static initialization and construction in .NET follows quite strict rules, cases of cyclic dependencies are not straightforward (see https://stackoverflow.com/questions/3681055/is-the-order-of-static-class-initialization-in-c-sharp-deterministic for such an example), so I would recommend to avoid them.

In fact the main contribution from my patch is that it avoids any static initialization/construction of `SchemaRegistry`.

If you want to see what is the main difference between 068f158830d4aa2e61862ebe10fc161e69d1f546 and my patch, we should investigate what happens when doing `ValidationOptions.Default` (and to actually check that everything is fine, we should also check `SchemaRegistry.Global`). This is going to be somewhat long to explin in detail, so I will split it into several comments 😇

ranma42:
In 068f158830d4aa2e61862ebe10fc161e69d1f546, when we access `ValidationOptions.Default`:
 - we are accessing `ValidationOptions`, hence the class must be initialized; the runtime executes the initializer https://github.com/gregsdennis/json-everything/blob/068f158830d4aa2e61862ebe10fc161e69d1f546/JsonSchema/ValidationOptions.cs#L19 which invokes the constructor https://github.com/gregsdennis/json-everything/blob/068f158830d4aa2e61862ebe10fc161e69d1f546/JsonSchema/ValidationOptions.cs#L95-L98 which invokes the class constuctor for `SchemaRegistry` at line 97; note that this corresponds to the stacktrace reported in the issue:
```
   at Json.Schema.SchemaRegistry..ctor(ValidationOptions options)
   at Json.Schema.ValidationOptions..ctor()
   at Json.Schema.ValidationOptions..cctor()
```
 - in order to run the `SchemaRegistry` instance constructor, the runtime starts the class initialization for `SchemaRegistry`, i.e. the static initializer https://github.com/gregsdennis/json-everything/blob/068f158830d4aa2e61862ebe10fc161e69d1f546/JsonSchema/SchemaRegistry.cs#L32 and class constructor https://github.com/gregsdennis/json-everything/blob/068f158830d4aa2e61862ebe10fc161e69d1f546/JsonSchema/SchemaRegistry.cs#L54-L58

At line 56 the code accesses `ValidationOptions.Default`, which is still `null` (the initializer has not yet completed!); this is exactly an example of the recommendation mentioned in the linked stackoverflow page:
> 17.4.5: It is possible for static fields with variable initializers to be observed in their default value state. However, this is strongly discouraged as a matter of style.

Since `ValidationOptions.Default` is still `null`, `Global = new SchemaRegistry(ValidationOptions.Default)` causes the construction of the global instance with a `_options` set to `null`, which then fails when `RegisterSubschemasAndGetBaseUri` accesses `ValidatingAs => _options.ValidatingAs`.

ranma42:
In my patch, when we access `ValidationOptions.Default`:
 - we are accessing `ValidationOptions`, hence the class must be initialized; the runtime executes the initializer https://github.com/ranma42/json-everything/blob/26231a9d26ada75be6df2786d26402382d48a27b/JsonSchema/ValidationOptions.cs#L19 which invokes the constructor https://github.com/ranma42/json-everything/blob/26231a9d26ada75be6df2786d26402382d48a27b/JsonSchema/ValidationOptions.cs#L98-L101 which invokes the class constructor for `SchemaRegistry` at line 97; so far everything matches what was going on in 068f158830d4aa2e61862ebe10fc161e69d1f546
 - in order to run the `SchemaRegistry` instance constructor, the runtime starts the class initialization for `SchemaRegistry`, i.e. the static initializer https://github.com/ranma42/json-everything/blob/26231a9d26ada75be6df2786d26402382d48a27b/JsonSchema/SchemaRegistry.cs#L30; note that there is no class constructor to invoke after my patch
 - the `SchemaRegistry` instance constructor https://github.com/ranma42/json-everything/blob/26231a9d26ada75be6df2786d26402382d48a27b/JsonSchema/SchemaRegistry.cs#L53-L56 can now continue with the execution, setting `_options` to its argument (the `ValidationOptions` that is about to be set as `Default`)
 - the `ValidationOptions` instance constructor https://github.com/ranma42/json-everything/blob/26231a9d26ada75be6df2786d26402382d48a27b/JsonSchema/ValidationOptions.cs#L98-L101 can now continue, setting the `SchemaRegistry` property to the newly-created `SchemaRegistry` instance
 - the initialization for `ValidationOptions.Default` can finally complete (note that at this point `ValidationOptions.Default` is initialized and `SchemaRegistry.Global` can be safely invoked as well) and now the `ValidationOptions` class constructor https://github.com/ranma42/json-everything/blob/26231a9d26ada75be6df2786d26402382d48a27b/JsonSchema/ValidationOptions.cs#L103-L106 runs, which results in the initialization of the metaschemas
 - finally the caller receives the result for `ValidationOptions.Default`

Hmmm... I thought that the issue with the non-embedding of the code snippets was because of an overlong comment, but now I think that they are so because they are from a different repo 😞

ranma42:
I believe that the original issue managed to stay somewhat hidden because in 068f158830d4aa2e61862ebe10fc161e69d1f546, when we access `SchemaRegistry.Global` the behavior is quite different:
 - we are accessing `SchemaRegistry`, hence the class must be initialized; the runtime executes the initializer https://github.com/gregsdennis/json-everything/blob/068f158830d4aa2e61862ebe10fc161e69d1f546/JsonSchema/SchemaRegistry.cs#L32 and class constructor https://github.com/gregsdennis/json-everything/blob/068f158830d4aa2e61862ebe10fc161e69d1f546/JsonSchema/SchemaRegistry.cs#L54-L58
At line 56 the code accesses `ValidationOptions.Default` (again!), but in this case the runtime has not yet started the initialization of `ValidationOptions`, hence...
 - the runtime executes the initializer https://github.com/gregsdennis/json-everything/blob/068f158830d4aa2e61862ebe10fc161e69d1f546/JsonSchema/ValidationOptions.cs#L19 which invokes the constructor https://github.com/gregsdennis/json-everything/blob/068f158830d4aa2e61862ebe10fc161e69d1f546/JsonSchema/ValidationOptions.cs#L95-L98 which invokes the instance constructor for `SchemaRegistry` at line 97
 - since the class initialization for `SchemaRegistry` is already ongoing, the runtime proceeds with the `SchemaRegistry` instance constructor https://github.com/gregsdennis/json-everything/blob/068f158830d4aa2e61862ebe10fc161e69d1f546/JsonSchema/SchemaRegistry.cs#L81-L84 and sets `_options` to its argument (the `ValidationOptions` that is about to be set as `Default`)
 - the `ValidationOptions` instance constructor https://github.com/gregsdennis/json-everything/blob/068f158830d4aa2e61862ebe10fc161e69d1f546/JsonSchema/ValidationOptions.cs#L95-L98 can now continue, setting the `SchemaRegistry` property to the newly-created `SchemaRegistry` instance
 - the initialization for `ValidationOptions.Default` can complete and the `ValidationOptions.Default` value becomes available for the `SchemaRegistry` class constructor https://github.com/gregsdennis/json-everything/blob/068f158830d4aa2e61862ebe10fc161e69d1f546/JsonSchema/SchemaRegistry.cs#L56 which can proceed with the metaschema registration
 - finally (after the class constructor completes) the caller receives the value of `SchemaRegistry.Global`

ranma42:
As you can see the `SchemaRegistry.Global`-driven initialization works just fine in 068f158830d4aa2e61862ebe10fc161e69d1f546, but if the initialization starts from `ValidationOptions.Default`, the runtime will always fail to initialize the classes properly.

Instead, my patch:
 - changes the code structure so that the `ValidationOptions.Default` works
 - ensures that the `SchemaRegistry.Global` initialization is basically a no-op and immediately defers to the `ValidationOptions.Default` initialization, because `SchemaRegistry` does no static initialization/construction (except for the `_empty` thing) and relies on `public static SchemaRegistry Global => ValidationOptions.Default.SchemaRegistry;`

If there are some changes in the patch that are unclear or if the patch is causing regressions, I would be happy to work some more on it (and/or on these comments) in the next few days :)

gregsdennis:
Yeah, it looks like all the potential loops are set up right.  I never really liked how I had it.  This seems simpler.

I've changed the target branch so I can do some Nuget admin stuff.

ranma42:
Cool, the patch is already available through nuget! Thank you :)

PR #185, opened 2021-11-15
Issue184 array support for pointers

### LINKED ISSUE #184: Add array navigation support to JsonPointer
The existing `JsonPointer.Create<T>` implementation has code for "usual" `IReadOnlyList`s:

https://github.com/gregsdennis/json-everything/blob/62d55e8b3c6537cf9dce99d81ccb336e2f88f3c5/JsonPointer/JsonPointer.cs#L249-L256

However, arrays are a special case and need different treatment; I think it would look something like this:

```csharp
if (
    body is BinaryExpression binaryExpression and { NodeType: ExpressionType.ArrayIndex }
    && binaryExpression.Right is ConstantExpression arrayIndexConstantExpression
)
{
    // Array index
    segments.Insert(0, PointerSegment.Create(arrayIndexConstantExpression.Value.ToString())); 
    body = binaryExpression.Left;
}
```

gregsdennis:
Resolves #184

### DIFF
--- a/JsonPointer/JsonPointer.cs
+++ b/JsonPointer/JsonPointer.cs
@@ -253,6 +253,17 @@ namespace Json.Pointer
 				{
 					segments.Insert(0, PointerSegment.Create(mce.Arguments[0].ToString()));
 					body = mce.Object;
+				} else if (body is BinaryExpression
+				           {
+					           Right: ConstantExpression arrayIndexConstantExpression
+				           } binaryExpression and
+				           {
+					           NodeType: ExpressionType.ArrayIndex
+				           })
+				{
+					// Array index
+					segments.Insert(0, PointerSegment.Create(arrayIndexConstantExpression.Value.ToString()));
+					body = binaryExpression.Left;
 				}
 				else if (body is ParameterExpression) break; // this is the param of the expression itself.
 				else throw new NotSupportedException($"Expression nodes of type {body.NodeType} are not currently supported.");
--- a/JsonPointer/JsonPointer.cs
+++ b/JsonPointer/JsonPointer.cs
@@ -253,16 +253,11 @@ namespace Json.Pointer
 				{
 					segments.Insert(0, PointerSegment.Create(mce.Arguments[0].ToString()));
 					body = mce.Object;
-				} else if (body is BinaryExpression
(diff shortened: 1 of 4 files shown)


JsonPointer/JsonPointer.cs
```diff
@@ -253,6 +253,17 @@ public static JsonPointer Create<T>(Expression<Func<T, object>> expression)
 				{
 					segments.Insert(0, PointerSegment.Create(mce.Arguments[0].ToString()));
 					body = mce.Object;
+				} else if (body is BinaryExpression
+				           {
+					           Right: ConstantExpression arrayIndexConstantExpression
+				           } binaryExpression and
+				           {
+					           NodeType: ExpressionType.ArrayIndex
+				           })
```
gregsdennis, line 262: This is so weird.  I don't like formatting `if`s on multiple lines unless each line is a sequence of conditionals, but I'm not sure how else do this nicely.

Arithmomaniac: I would 
(a) be less strict about having the brackets on multiple lines (most examples of pattern matching don't do this anyways), and 
(b) put each `and` on its own line. 

Like 
```csharp
body is BinaryExpression { Right: ConstantExpression arrayIndexConstantExpression }
    and { NodeType: ExpressionType.ArrayIndex  }
```

gregsdennis: There's actually a variable declaration in there, too, so it'd be

```c#
body is BinaryExpression { Right: ConstantExpression arrayIndexConstantExpression } binaryExpression
    and { NodeType: ExpressionType.ArrayIndex  }
```

I tend to like `&&` at the end of lines, but `and` at the end is too long. 🤷‍♂️

gregsdennis: Yeah, I think that's the best I'm going to do.

Arithmomaniac: I know you completed this PR already, but there's this alternative for a future PR:
```csharp
body is BinaryExpression {
    Right: ConstantExpression arrayIndexConstantExpression,
    NodeType: ExpressionType.ArrayIndex 
} binaryExpression
```

PR #187, opened 2021-11-25
Added a OneOf Intent

dazerdude:
I was coding and found this omission in the generation api. I basically copied AnyOfIntent.

### DIFF
--- a/JsonSchema.Generation/Intents/OneOfIntent.cs
+++ b/JsonSchema.Generation/Intents/OneOfIntent.cs
@@ -1,0 +1,76 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Json.Pointer;
+
+namespace Json.Schema.Generation.Intents
+{
+	/// <summary>
+	/// Provides intent to create a `oneOf` keyword.
+	/// </summary>
+	public class OneOfIntent : ISchemaKeywordIntent
+	{
+		/// <summary>
+		/// Gets the subschemas to include.
+		/// </summary>
+		public List<IEnumerable<ISchemaKeywordIntent>> Subschemas { get; }
+
+		/// <summary>
+		/// Creates a new instance of the <see cref="OneOfIntent"/> class.
+		/// </summary>
+		/// <param name="subschemas">The subschemas to include.</param>
+		public OneOfIntent(IEnumerable<IEnumerable<ISchemaKeywordIntent>> subschemas)
+		{
+			Subschemas = subschemas.ToList();
+		}
+
+		/// <summary>
+		/// Creates a new instance of the <see cref="OneOfIntent"/> class.
+		/// </summary>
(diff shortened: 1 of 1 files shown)


gregsdennis:
How would this be used?  Just having the intent doesn't automatically use it.

For example, the `AnyOfIntent` is used in the `JsonNumberHandlingAttributeHandler`, which allows for numbers to be incoded in either numbers or strings, including the literals, `NaN`, `PositiveInfinity`, and `NegativeInfinity`.

dazerdude:
I'm using it for my own implementations of ISchemaGenerator. I found it odd that such a core intent wasn't available in the public api, so I figured I'd submit it back. If you don't want it, that's fine, but I'm happy to write some tests if you do.

The specific use case I'm using it for is abstract base types that can be one of several concrete subtypes.

gregsdennis:
I'm interested in your use case and how you're signalling to the system that a `oneOf` should be created.

dazerdude:
This is not exactly my code, but it's the same idea. It's similar to the way my json serializer picks the right subtype to deserialize. TypeMappings is prepoulated which only works because I have a static set of assemblies where I know the concrete subtypes will be.

`
public class ConcreteSchemaGenerator : ISchemaGenerator
{
    public List<Dictionary<Type, List<Type>>> TypeMappings; // [Abstract base type -> [Concrete subtypes]]
    public bool Handles(Type type) => TypeMappings.ContainsKey(type);
    public void AddContratints(Context ctx)
    {
        var subtypes = TypeMappings[ctx.Type];
        var subschemas = subtypes.Select(t => ContextCache(t, ctx.Attrs, ctx.Config))
        ctx.Intents.Add(new OneOfIntent(subschemas));
    }
}
`

gregsdennis:
Okay... See, _that's_ the feature that should be added: supporting multiple implementations of an abstract type.  Though I might still consider that an `anyOf` in case there's overlap between the implementations.

dazerdude:
Okay, well my solution to that is not really generalizable.

Tbh, as an outsider approaching the library, I'd not expect a feature like that. I'd basically expect it to have the same general level of abstraction and features as System.Text.Json, it definitely can't handle that out of the box.

gregsdennis:
I think the registry is the way to go.  The library needs to know what options there are.  Also, I wouldn't want it to try to list implementations of things like `IComparable`.

dazerdude:
A pitfall to be aware of, is that if you get too many concrete types for a single subtype, they really become *really* slow to compute with the generic anyOf/oneOf keywords you have. I had to build an optimized oneOf keyword to pull the value from the $type property that I require my json objects to have. It then used this value to just validate the specific relevant subschema (since all the others have different values for the $type property).

There is probably a generic version of this optimization you can make if you do add this feature, and you'll probably need it to be at all performant. I was hitting perf issues past like 5 subtypes.

PR #195, opened 2021-12-04
Issue193 custom format behaviors

### LINKED ISSUE #193: handling of unknown formats
**Describe the bug**
https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.7.2.3

As per the spec, an implementation must not fail to collect unknown formats as annotations.

There is also the bit in there about format assertion vocabulary, is that the same as specifying `RequireFormatValidation`?

Currently it seems that there is no way to force a failure when encountering an unknown format.

**To Reproduce**
Define a schema with an unknown format. When validating it will return an annotation but the value will be null.

**Expected behavior**
The annotation value should contain the name of the unknown format.

If `RequireFormatValidation` is set to true, additionally I might expect the validation to fail which it currently does not.

gregsdennis:
Resolves #193


gregsdennis:
I added .net 6 to the test project, but the build doesn't support it yet.

PR #200, opened 2021-12-12
JsonSchema.Net v2.0.0

### LINKED ISSUE #197: Json Schema allocates a lot
**Environment**
- OS: Windows
- .Net Target: 6

**Describe the bug**
Json Schema allocates a _lot_ when validating reasonably complex json files. For my specific case it appears to allocate ~12 GB.

**To Reproduce**
My json files aren't too complex, but they are somewhat deep. They top out around 24 levels deep. I do have a fair bit of oneOf/anyOfs and an object that contain a list of itself.

**Expected behavior**
Processing a json file shouldn't really take that much ram. Especially when I have optimizations enabled.

**Additional context**
The perf analyzer basically says they're all Validation Contexts and associated structures. I ripped these out of my fork, and just use the one context everywhere. It works great for me, and gets the job done in a constant ~50MB, but I disabled some features/keywords that I wasn't using to get it done. Additional substate is pushed into this master context (and pulled out when the keyword exits).
https://github.com/gregsdennis/json-everything/blob/master/JsonSchema/ValidationContext.cs#L146

Some potentially low hanging fruit that could help:
NavigatedReferences creates a HashSet when traveling down a level but staying with the same instance? Navigating to the same reference for the same object (even from a different keywords) seems unnecessary? I don't know all the details here, but in my case these collections were growing quite large and pinning it to the json element instead of the nested context worked great. I have tons of references but only in document.
https://github.com/gregsdennis/json-everything/blob/24f2557ce2ec3a146b6047f55b1f486bd024c0f8/JsonSchema/ValidationContext.cs#L167

The Clone call can also allocate some, although it presumably only clones the relevant parts of the document, every sublevel clones it's parent even if the sublevel doesn't reference a different part of the document. Passing JsonElements by reference (instead of embedding them in the context) helped. Cloning once at top level so the reference the document isn't lost (or just taking the JsonDocument in the api, potentially just two entry points).
https://github.com/gregsdennis/json-everything/blob/24f2557ce2ec3a146b6047f55b1f486bd024c0f8/JsonSchema/ValidationContext.cs#L159

I removed the subschema location pointer entirely since I didn't really care. InstanceLocation is quite useful for debugging, but I decomposed it into a Stack of strings held by the master context, for my current document location, and then dynamically resolved that when I needed it (for final error points and the like). The stack is popped on the way up. This saves a fair bit of space and allocation time. Perhaps a list of PointerSegments but even that strikes me as a bit unnecessary as it greedily copies the string.

In the same vein, you have a lot of format strings of the form $"{i}" or $"{name}". I could be wrong, and these could be optimized away, but I believe these are translated to calls to string.Format . Making them relatively much slower than just calling ToString on the object (or just passing it in if it's a string like JsonProperty.Name). I don't think they were contributing too much to the memory issue though as I believe the runtime won't allocate a new string if the output string already exists (but I'm not totally sure).
https://github.com/gregsdennis/json-everything/blob/24f2557ce2ec3a146b6047f55b1f486bd024c0f8/JsonSchema/PatternPropertiesKeyword.cs#L87
https://github.com/gregsdennis/json-everything/blob/24f2557ce2ec3a146b6047f55b1f486bd024c0f8/JsonSchema/ContainsKeyword.cs#L63

gregsdennis:
Depends on #201 

Resolves #197

### DIFF
--- a/JsonPath.Tests/Suite/CburgmerFeatureValidationTests.cs
+++ b/JsonPath.Tests/Suite/CburgmerFeatureValidationTests.cs
@@ -28,6 +28,7 @@ namespace Json.Path.Tests.Suite
 			{
 				// expect these to be out of spec soon
 				"$.key-dash",
+				"$.-1",
 				"$.length",
 
 				// big numbers not supported
--- a/JsonSchema.Data/DataKeyword.cs
+++ b/JsonSchema.Data/DataKeyword.cs
@@ -75,8 +75,7 @@ namespace Json.Schema.Data
 			}
 			catch (JsonException e)
 			{
-				context.IsValid = false;
-				context.Message = e.Message;
+				context.LocalResult.Fail(e.Message);
 				return;
 			}
 
@@ -90,12 +89,12 @@ namespace Json.Schema.Data
 			var baseUri = parts[0];
 			var fragment = parts.Length > 1 ? parts[1] : null;
 
-			JsonElement data = default;
+			JsonElement data;
 			if (!string.IsNullOrEmpty(baseUri))
 			{
 				if (Uri.TryCreate(baseUri, UriKind.Absolute, out var newUri))
 					data = Download(newUri);
-				else if (context.CurrentUri != null)
+				else
 				{
 					var uriFolder = context.CurrentUri.OriginalString.EndsWith("/")
 						? context.CurrentUri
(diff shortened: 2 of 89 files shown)


gregsdennis:
Test failures are in JSON Logic.  I'll address that separately.

PR #204, opened 2022-01-12
unknown format validations should not error

### LINKED ISSUE #203: JSON validation regression in JsonSchema.Net v1.12.0
**Environment**
- Nuget Version: v1.12.0
- OS: Windows Server 2022, Ubuntu 20.04 and Mac OS 11
- .Net Target: .NET 6

**Describe the bug**
Dependabot bumped this dependency version in one of my projects and a bunch of tests failed. I've checked it out locally and it appears there was a regression introduced between v1.11.6 and v1.12.0. But I can't tell what might be causing that by looking at the changes.

**To Reproduce**
I have tried to create a fairly minimal reproduction here https://github.com/coderpatros/jsonschema.net-validation-repro

That has the version set as v1.11.6 so you can confirm it was working as expected first.

**Expected behavior**
I expect the example in the repro to validate as valid JSON.

gregsdennis:
Fixes #203 
Related to #193

### DIFF
--- a/JsonSchema.Tests/FormatTests.cs
+++ b/JsonSchema.Tests/FormatTests.cs
@@ -73,7 +73,11 @@ namespace Json.Schema.Tests
 		[Test]
 		public void UnknownFormat_Assertion_FailsValidation()
 		{
-			var options = new ValidationOptions {OutputFormat = OutputFormat.Detailed};
+			var options = new ValidationOptions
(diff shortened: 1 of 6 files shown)


gregsdennis:
Those test failures are from JSON Logic - unrelated to this change.  I'll deal with that separately.

Issue #221, opened 2022-02-21
All values fail against the false schema when additionalProperties is set to false


fredjeck:
**Environment**
- Nuget Package: . JsonSchema.Net
- Nuget Version: 2.0.3
- OS: Linux & Windows
- .Net Target: 6

**Describe the bug**
First and before : Kudos for the awesome job 😍
Potentially related to #216 
When validating a faulty schema with _additionalProperties_ set to false, validation yields extraneous error messages ("All values fail against the false schema") those extraenous messages are not present when _additionalProperties_ is set to true

**To Reproduce**
Schema
```json
{
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "type": "object",
    "additionalProperties": false,
    "properties": {
        "firstname": {"type": "string"},
        "phone": {"type": "string"},
        "email": {"$ref": "#/$defs/emailAddress"}
    },
    "required": ["firstname","phone","email"],
    "$defs": {
        "emailAddress": {
                "type": "object",
                "additionalProperties": false,
                "properties": {
                    "email": {"type": "string"},
                    "isConfirmed": {"type": "boolean"}
                },
                "required": [ "email","isConfirmed"]
        }
    }
}
```

The following sample is missing the _isConfirmed_ property
```json
{
    "firstname": "firstnameTest",
    "phone": "123456",
    "email": {
        "email": "test@test.com"
    }
}
```

This generates the following validation errors :

- Required properties [isConfirmed] were not present found at #/email
- All values fail against the false schema found at #/firstname
- All values fail against the false schema found at #/phone
- All values fail against the false schema found at #/email

Setting _additionalProperties_ is set to true only one validation error is raised 

- Required properties [isConfirmed] were not present found at #/email

**Expected behavior**
Behavior should be consistent independently from the _additionalProperty_ property and should only report the first error

gregsdennis:
I don't see how this is a bug.  A `true` schema passes all values, so no errors would arise.

The way the spec is worded, any properties not validated by another "properties" keyword are validated against the subschema in `additionalProperties`.  This means that these errors should be present when that subschema is false because those keywords are failing the `false` schema.

We are currently looking at removing the "annotation" behavior that drives this and returning to the draft 7 language where the keyword instead looks at its sibling keywords directly to determine which properties to assess.

fredjeck:
[...]

Issue #216, opened 2022-02-11
All values fail against the false schema issue


joyse1:
**Environment**
Nuget Package: JsonSchema.Net
Nuget Version: 2.0.2
OS: Win10
.Net Target: .Net Framework 4.7.2

**Describe the bug**
Schema validation, when json data is incorrect ( has additional non unique item in Contacts Array )  reports additional validation messages " All values fail against the false schema ". Validation is passing when ununique contact item is removed.

**To Reproduce**
_Json Schema:_
```json
{
	"$schema": "http://json-schema.org/draft-07/schema#",
	"$id": "http://messagebroker.fff.pl/rpn/dci/kkt.json",
	"type": "object",
	"title": "JSON Schema dokumentu bazowego zdarzenia RPN-DCI-KKT. Version 3b. 2020-01-21",
	"definitions": {
		"guid": {
			"$id": "#guid",
			"title": "Definicja obligatoryjnego GUID (regex)",
			"type": "string",
			"examples": [
				"09C9A8DA-B40F-4E3A-9746-7B10AFEC2C4F",
				"2d9bbb00-878f-49b0-9d48-767ce3e12dee"
			],
			"minLength": 36,
			"maxLength": 36,
			"pattern": "^([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})$"
		},
		"string50Nullable": {
			"$id": "#string50Nullable",
			"title": "Definicja opcjonalnego tekstu o długości nie przekraczającej 50 znaków",
			"default": null,
			"anyOf": [
				{
					"type": "string",
					"minLength": 1,
					"maxLength": 50,
					"pattern": "^(.*)$"
				},
				{
					"type": "null"
				}
			]
		},
		"string250Nullable": {
			"$id": "#string250Nullable",
			"title": "Definicja opcjonalnego tekstu o długości nie przekraczającej 250 znaków",
			"default": null,
			"anyOf": [
				{
					"type": "string",
					"minLength": 1,
					"maxLength": 250,
					"pattern": "^(.*)$"
				},
				{
					"type": "null"
				}
			]
		},
		"dictionaryElement": {
			"$id": "#dictionaryElement",
			"type": "object",
			"title": "Definicja elementu słownika",
			"additionalProperties": false,
			"required": [
				"ElementCode",
				"ElementName"
			],
			"properties": {
				"ElementCode": {
[...]

Issue #89, opened 2021-03-22
JsonSchema errors for additionalProperties not the same as 2019-09 draft specification


brigadier90:
**.net core 3.1**
- Nuget Package: JsonSchema.Net
- Nuget Version: 1.9.3
- OS: Windows 10 version 20H2

I am trying to replicate the scenario specified in this official [draft](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.10.4). Here is my code:

```


using Json.Pointer;
using Json.Schema;
using System;
using System.Text.Json;

namespace ConsoleApp1
{
    class Program
    {
        static void GetErrorString(
            ValidationResults results,
            ref string errorString)
        {
            if (results.Message != null)
            {
                errorString += results.InstanceLocation + ": ";
                errorString += results.Message + "\n";
            }

            foreach (var result in results.NestedResults)
                GetErrorString(result, ref errorString);
        }

        static void Main(string[] args)
        {
            var schema = JsonSchema.FromFile("./point.json");
            var jsnStr = @"[
                {""x"": 2.5, ""y"":1.3}, 
                {""x"": 1, ""z"":6.7}
            ]";

            var jsnObj = JsonDocument.Parse(jsnStr).RootElement;

            ValidationOptions options = new ValidationOptions();
            options.OutputFormat = OutputFormat.Detailed;
            options.ValidateAs = Draft.Draft201909;

            var result = schema.Validate(jsnObj, options);

            var errorString = "";
            if(!result.IsValid)
                GetErrorString(result, ref errorString);

            Console.WriteLine(errorString);
        }
    }
}
```
This is my output:
```
#: Value has less than 3 items
#/1: Required properties [y] were not present
#/1/z: All values fail against the false schema
```
I noticed the error for pointer "#/1/z" is different to the one specified in the draft [here](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.10.4.3) which says: "Additional property 'z' found but was invalid". 

This becomes especially problematic when I have the following schema:
```
{
  "$schema": "https://json-schema.org/draft/2019-09/schema",
  "$id": "/CheckRequest",
  "type": "object",
  "properties": {
    "code": {
      "type": "string",
      "minLength": 4,
      "maxLength": 10
    },
    "phone": {
      "type": "string",
      "pattern": "^\\+[1-9]\\d{1,14}$"
    }
  },
  "required": [ "code", "phone"],
  "additionalProperties":  false 
}
```
[...]

Issue #39, opened 2020-11-18
"additionalProperties" being validated as a property 


nabinked:
**Describe the bug**
The `additionalProperties` keyword is being treated as a property.

**To Reproduce**
```csharp
using System;
using System.Text.Json;
using Json.Schema;

var schema = @"{
  ""$schema"": ""http://json-schema.org/draft-07/schema#"",
  ""additionalProperties"": false
}";

var data = @"{
  ""prop1"":""val""
}";

var jsonSchema = JsonSchema.FromText(schema);
var validationResult = jsonSchema.Validate(JsonDocument.Parse(data).RootElement, new ValidationOptions { OutputFormat = OutputFormat.Basic });
Console.WriteLine(JsonSerializer.Serialize(validationResult));

//outputs
{
    "valid": false,
    "keywordLocation": "#/additionalProperties/$false",
    "instanceLocation": "#/prop1",
    "error": "All values fail against the false schema"
}

```

**Expected behaviour**
The error should indicate that additional properties is present.
![image](https://user-images.githubusercontent.com/7703179/99497150-9c288a80-29c9-11eb-859c-a6962d481e44.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
- .NET 5 (SDK 5.0.100)
 - Nuget Version [e.g. 22]
![image](https://user-images.githubusercontent.com/7703179/99497515-25d85800-29ca-11eb-8738-5e786ccebd37.png)

gregsdennis:
The schema *is* indicating that.  Perhaps the confusion is that the error is technically coming from the `false` subschema of the `additionalProperties`.  Regardless, the instance fails for the same reason.

nabinked:
Yes, basically the error message is a bit misleading. Please feel free to update the issue to reflect the actual issue here.

gregsdennis:
The error message this library gives is more accurate.  It's the subschema of the `additionalProperties` that's generating the error, not the keyword itself.  Additionally Newtonsoft hasn't updated their output formatting to 2019-09 yet; it just gives an error message.

nabinked:
Also I am not sure if this is expected but should the 2nd error be present in the errors list ?
**schema**
```json
{
  "$schema": "http://json-schema.org/draft-06/schema#",
  "additionalProperties": false,
  "properties": {
    "firstName": {
      "const":"foo"
    }
  }
}
```
**data**
```json
{
  "firstName": "bar"
}
```
**outputs**
```json
{
  "valid": false,
[...]

PR #232, opened 2022-03-03
adjust how annotations are used (discussion in JSON Schema spec for 2020-12 patch)

gregsdennis:
Depends on https://github.com/json-schema-org/json-schema-spec/issues/1172

Relates to:
- #39 
- #89 
- #216 
- #221


gregsdennis:
I'm just going to merge this.  We (JSON Schema) have agreed that for 2020-12, the re-interpretation is up to the implementation and we're not going to make any functional changes.

PR #236, opened 2022-03-22
private JetBrains.Annotations dependency

WeihanLi:
Let the `JetBrains.Annotations` dependency be private, it can be private so that won't be an external dependency

https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,6 @@
+<Project>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All"/>
+    <PackageReference Include="JetBrains.Annotations" Version="2021.2.0" PrivateAssets="All" />
+  </ItemGroup>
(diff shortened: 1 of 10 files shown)


gregsdennis:
Thanks for the suggestion, but this dependency needs to remain public.  I have various R# attributes placed specifically for clients to consume, like the [RegexPattern](https://resharper-support.jetbrains.com/hc/en-us/community/posts/115000242290-RegexPatternAttribute-in-an-attribute-argument) attribute.

Out of curiosity, is this actually causing a problem for you, or is this just a philosophical "reduce dependencies" suggestion?

WeihanLi:
Hi @gregsdennis , no problem met so far, just think it should be a development-only dependency, and I try to build the whole project, it would not fail

gregsdennis:
Okay. Thank you. I hear you, but I have my reasons for it being the way it is as explained above.

PR #238, opened 2022-03-23
Add JsonSchemaExtensions

### LINKED ISSUE #237: Could we introduce extension methods for validating `JsonDocument`/json string by json schema
**Environment**
- Nuget Version: 2.1.2
- OS: Windows/Linux
- .Net Target: .NET 6

Maybe we could introduce some extensions to make it more easier, for example:

``` c#
public static ValidationResults Validate(this JsonSchema jsonSchema, JsonDocument jsonDocument, ValidationOptions validationOptions = null)
{
    return jsonSchema.Validate(jsonDocument.RootElement, validationOptions);
}

public static ValidationResults Validate(this JsonSchema jsonSchema, string jsonString, ValidationOptions validationOptions = null)
{
    using var jsonDocument = JsonDocument.Parse(jsonString);
    return jsonSchema.Validate(jsonDocument, validationOptions);
}
```

**Describe alternatives you've considered**

Not sure if it's reasonable and valuable to add the `JsonDocument` method as an instance method

**Additional context**

I'm new here, maybe there're some magic ways that I don't know

WeihanLi:
Fixes #237

### DIFF
--- a/JsonSchema/JsonSchemaExtensions.cs
+++ b/JsonSchema/JsonSchemaExtensions.cs
@@ -1,0 +1,35 @@
+﻿using System.Text.Json;
+
+namespace Json.Schema
+{
+	/// <summary>
+	/// Some extensions for <see cref="JsonSchema"/>
+	/// </summary>
+	public static class JsonSchemaExtensions
+	{
+		/// <summary>
+		/// Validate JsonDocument with jsonSchema
+		/// </summary>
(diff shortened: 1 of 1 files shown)


gregsdennis (approved):

gregsdennis:
Thanks for this.  I'm going to edit the comments a bit, but the implementation looks good.

PR #249, opened 2022-04-04
Schema/configurable error messages

### LINKED ISSUE #190: Support for i18n and custom error messages
**Environment**
- Nuget Version: 1.11.6
- OS: Linux
- .Net Target: 5.x

**How can I help you today?**
Does this library support other languages for the error messages or provide a way to provide custom error messages when certain validation fails?

gregsdennis:
Resolves #190


gregsdennis:
Only way to check if the localization publishes is to run it from master, so... let's merge... 🤞

Issue #202, opened 2021-12-19
Json.Schema.Generation: schema from "legacy" immutable classes (before C# record)


ugumba:
Legacy immutable classes typically have getter-only properties, which are only settable by the constructor.

System.Text.Json automatically detects and supports these properties - it can also be enforced by [JsonConstructor].
Getter-only properties are, however, filtered out and ignored by ObjectSchemaGenerator.
It's a bit jarring when Json.Schema.Generation is missing semantics from System.Text.Json.

A couple of workarounds for when the code is available, in order of personal preference:

* Port to C# record (the init-only getter seems to get past the filter in ObjectSchemaGenerator)
* Use readonly field instead of getter-only property

When types come from 3rd party assemblies, however, these workarounds are not available.

gregsdennis:
I mentioned the reasoning behind the current support of only read/write properties [here](https://github.com/gregsdennis/json-everything/issues/149#issuecomment-892492510).

It seems this may have changed recently with .net 6.

Also, I'm working on a [2.0](https://github.com/gregsdennis/json-everything/issues/198).  I welcome your suggestions there.

ugumba:
Thanks!
Just to note that this is not new with .NET 6 - System.Text.Json has handled readonly properties, matching them up with constructor parameters at least since .NET 5.  I know, it's a somewhat annoying bit of magic for J.S.Generation to consider mirroring.

gregsdennis:
Version 2 is out.  It now supports single-accessor properties and readonly fields by adding the appropriate keywords to the schema.  This behavior can be overridden by the attributes.

Issue #149, opened 2021-08-04
How to serialize generated schemas into JSON Schemas


joelhoisko:
**Environment**
- Nuget Package: `JsonSchema.Net 1.10.7`, `JsonSchema.Net.Generation 1.8.0`
- Nuget Version: NuGet Command Line 5.10.0.7240
- OS: Ubuntu 20.04
- .Net Target: .Net 5

I've read through the docs, and I couldn't find an example on how to serialize the generated `JsonSchema` object into a JSON Schema string. Looking at the test I saw that using `JsonSerializer.Serialize()` seemed to be the way to go about it. But the resulting JSON isn't a schema at all. Maybe I'm missing something, or that using attributes is mandatory, but[ reading the docs](https://gregsdennis.github.io/json-everything/usage/schema-generation.html#making-better-schemas) I thought that it would actually make a schema with properties for me without using attributes (the docs do say "maybe", not sure what that then means).

Example:
```csharp
using System;
using System.Text.Json;
using Json.Schema;
using Json.Schema.Generation;

namespace json_everything
{
    class Program
    {
        static void Main(string[] args)
        {
            var schema = new JsonSchemaBuilder().FromType(typeof(MyClass)).Build();
            var schemaJson = JsonSerializer.Serialize(schema, new JsonSerializerOptions { WriteIndented = true });
            Console.WriteLine(schemaJson);
        }
    }

    class MyClass
    {
        public MyClass(string myString)
        {
            MyString = myString;
        }
        public string MyString { get; }
    }
}

```

output:
```json
{
  "type": "object"
}
```

I was expecting to get something like:
```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "properties": {
    "MyClass": {
      "type": "object",
      "properties": {
        "MyString": {
          "type": "string"
        }
      },
      "required": [
        "MyString"
      ]
    }
  },
  "required": [
    "MyClass"
  ]
}
```

Is there a way to actually output a serialized JSON Schema from the `JsonSchema` class or am I missing something here?

joelhoisko:
Changing the `MyClass.MyString` to also have a setter made it a bit better:

```csharp
class MyClass
{
    public MyClass(string myString)
    {
        MyString = myString;
    }
    public string MyString { get; set; }
}
```

Output:
```json
{
  "type": "object",
  "properties": {
    "MyString": {
      "type": "string"
    }
  }
}
```

[...]

PR #261, opened 2022-05-09
Schema generation v2.0.0

### LINKED ISSUE #198: Schema Generation 2.0
This single issue is to consolidate other issues raised against this library.

- #117
  - Base class requirements could be extracted into their own definition.
  - Sealed classes should protect against new properties.
- #143
  - Make use of (or otherwise support the use of) .Net's data annotations.
- #149
  - Allow certain attributes on types so that they can be applied to the top-level schema.
  - Possibly support read-only and write-only properties (although not supported by `System.Text.Json` serialization).

### LINKED ISSUE #206: Incorrect XML-doc summaries on some intent constructors
**Environment**
- Nuget Version: 1.8.3
- OS: Windows
- .Net Target: .NET 6

**Describe the bug**

1.  The XML-doc summary for Json.Schema.Generation.Intents.DescriptionIntent constructor incorrectly states that it "Creates a new **MinimumIntent** instance."

See: https://github.com/gregsdennis/json-everything/blob/5b257ecb789d0f434269713182fe35282c921e89/JsonSchema.Generation/Intents/DescriptionIntent.cs#L14

2. The XML-doc summaries for Json.Schema.Generation.Intents.EnumIntent constructors appear to have been copied from the "Apply" method.

See: https://github.com/gregsdennis/json-everything/blob/5b257ecb789d0f434269713182fe35282c921e89/JsonSchema.Generation/Intents/EnumIntent.cs#L19
and: https://github.com/gregsdennis/json-everything/blob/5b257ecb789d0f434269713182fe35282c921e89/JsonSchema.Generation/Intents/EnumIntent.cs#L28

**To Reproduce**
Comments visible in IntelliSense when writing code to construct a new DescriptionIntent or EnumIntent.

**Expected behavior**
The summary should indicate that the constructor creates a new DescriptionIntent or EnumIntent instance per summaries on other Intent constructors.

### LINKED ISSUE #256: MaxLenght  Annotation
I have a class with annotation 
[Required]
and  [MaxLenth(100)

The Schema is generated ok for the required property but it ignores the MaxLenthtotally.
I even try Maximum and Minimum but the schema doesn't add this as items to be validated. 

Is there any configuration to be done in the schema generation to allow these annotations being picked up on the schema ?

gregsdennis:
Resolves #198
Resolves #206
Resolves #256

Related to #149


gregsdennis:
This also resolved #202

Issue #174, opened 2021-11-01
Extraneous JsonSchema.Net NuGet Dependencies?


ArcanoxDragon:
Currently, JsonSchema.Net.Generation has a direct dependency on the "Humanizer" metapackage, which pulls in all of their localization resources as "compile" dependencies if I add "JsonSchema.Net.Generation" to a project. This means I get a ton more `.dll`s in my output folder than without the Generation package. It looks like the only place Humanizer is used is when changing the casing of property names during schema generation; is there a chance the dependency could be changed to `Humanizer.Core`? I don't think their culture-specific localization packages are necessary for case changes on property names, but I'm not 100% sure on that. 

In addition, `JsonSchema.Net` itself has a strong dependency on `JetBrains.Annotations`, which caused my existing NuGet package references for that package to be updated in my entire solution, and all of the `bindingRedirects` (for every entry-level project) to need to be changed. I'm working in a .NET Framework solution with a ton of "entry" projects, so this is a bit undesirable. JetBrains suggests [in a blog post](https://blog.jetbrains.com/dotnet/2018/05/03/what-are-jetbrains-annotations/) that the annotations be included as source instead of as a `JetBrains.Annotations` dependency reference for libraries intended for consumption by others. They have additional documentation on that [here](
https://www.jetbrains.com/help/resharper/Code_Analysis__Annotations_in_Source_Code.html#embedding-declarations-of-code-annotations-in-your-source-code)

gregsdennis:
Thanks for the question.

I'll see what I can do about Humanizer, but the Jetbrains one needs to stay.

That blog post you reference is pretty old now, and they don't recommend copying the code in anymore.  In fact, in their options, you can see that they explicitly recommend the nuget package now.

![image](https://user-images.githubusercontent.com/2676804/139941626-960fd379-ee53-4dc8-8056-2b4671691361.png)

ArcanoxDragon:
Interesting; I've never noticed that blurb in the ReSharper settings now. Our code base has a lot of legacy Framework projects in it and, for historical reasons, all NuGet packages for `packages.config`-based projects are under version control, so I try and avoid needing to update things like `JetBrains.Annotations` when I can avoid it, but it looks like I may not be able to avoid it this time.

Hopefully the Humanizer dependency can indeed be lightened though; that will help a lot.

gregsdennis:
It looks like referencing Humanizer.Core is probably the preferred method for another package.  From their docs:

> Here are the options:
> 
[...]

PR #289, opened 2022-06-17
Remove JetBrains.Annotations reference from nuget package

jan-frost:
I would like to remove the reference to JetBrains.Annotations reference from the JsonPatch.Net nuget package.

According to https://blog.jetbrains.com/dotnet/2015/08/12/how-to-use-jetbrains-annotations-to-improve-resharper-inspections/, this is not the best practice, but I don't know, if you want to have the annotation source code in the repository. This change does not change with regards to the nuget package, as the annotations are not compiled into the assembly by default.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,7 @@
+<Project>
+  
+  <ItemGroup>
+    <PackageReference Include="JetBrains.Annotations" Version="2021.2.0" PrivateAssets="All" />
+  </ItemGroup>
+
+</Project>
--- a/JsonSchema/JsonSchema.csproj
+++ b/JsonSchema/JsonSchema.csproj
@@ -39,7 +39,6 @@
 	</ItemGroup>
 
 	<ItemGroup>
-		<PackageReference Include="JetBrains.Annotations" Version="2021.2.0" />
 		<PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
 	</ItemGroup>
 


gregsdennis:
Duplicate of #174.  Please have a read.

In short, I'm not removing this dependency.  That blog post is from 7 years ago, before they released the annotations as a package.

jan-frost:
Sorry, I should have had a look at the issues first. Thank you for the quick reply!

However, the current documentation of the latest Resharper release still basically says the same (see https://www.jetbrains.com/help/resharper/Code_Analysis__Annotations_in_Source_Code.html):

> By default, all annotation attribute classes are marked with [Conditional("JETBRAINS_ANNOTATIONS")] attribute so the compiler will ignore the attribute usages in your code, which means that no binary reference to the 'JetBrains.Annotations.dll' assembly will be produced. However, you can define 'JETBRAINS_ANNOTATIONS' conditional compilation symbol in your projects to preserve the attributes in the metadata.

So, unless I am misunderstanding this, if you do not set JETBRAINS_ANNOTATIONS in the build, the annotations will remain internal to the project and there is no need to have JetBrains.Annotations as a dependency to the nuget package.

Some background why this is important to me: I am working on a project in a highly regulated environment, which requires me to do a thorough OTS/SOUP analysis of anything we deploy with our software. That would currently include JetBrains.Annotations due to the transitive dependency in JsonPatch.Net.

gregsdennis:
Here's the history behind it:

I purposefully use these annotations in several places, specifically to help clients.

For example, in JsonSchema.Net, there is a `.Pattern()` extension method that takes a string. I've decorated this parameter with `[RegexPattern]` so that clients who also use R# will get RegEx syntax highlighting on the string they submit to this method.

Yes, I could define this locally in my own code, but then it creates a build conflict for anyone using the package in their code.  I dealt with this before with Manatee.Json.  The current setup is the suggested approach.

---

That said, JsonPatch.Net is getting the package transitively from JsonSchema.Net, which uses the attributes as indicated above.  I can't only remove it from JsonPatch.Net.

jan-frost:
Again, thank you for the reply! I hope I am not annoying you with the persistence, but I am really trying to understand how to use JetBrains.Annotations correctly here.

First of all: it is absolutely great that we have these annotations in the json-everything packages. However, as far as I can see, the attributes are not compiled into the assemblies, so there is really no need to add a public dependency to the JetBrains.Annotations package.

If you would want to expose the attributes in the nuget package, you would need to compile with the JETBRAINS_ANNOTATIONS switch. Or you add the source directly, but I really don't know how that behaves, if the nuget package is references as well.

-------------------

I did a quick test using the RegexFormat class. The decompiled binary does not have the attributes:

```csharp
using System;
using System.Text.Json;
using System.Text.RegularExpressions;


#nullable enable
namespace Json.Schema
{
  /// <summary>A regular-expression-based format.</summary>
  public class RegexFormat : Format
  {
    private readonly Regex _regex;

    /// <summary>
    /// Creates a new <see cref="T:Json.Schema.RegexFormat" />.
    /// </summary>
    /// <param name="key">The format key.</param>
    /// <param name="regex">The regular expression.</param>
    public RegexFormat(string key, string regex)
      : base(key)
    {
      this._regex = new Regex(regex, RegexOptions.Compiled | RegexOptions.ECMAScript);
    }

    /// <summary>Validates an instance against a format.</summary>
    /// <param name="element">The element to validate.</param>
    /// <returns><code>true</code> if the value is a match for the regular expression; <code>false</code> otherwise.</returns>
    [Obsolete("This method isn't called anymore.  It will be removed in the next major version.")]
    public override bool Validate(JsonElement element) => this.Validate(element, out string _);

    public override bool Validate(JsonElement element, out string? errorMessage)
    {
      if (((JsonElement) ref element).ValueKind != 3)
      {
        errorMessage = (string) null;
        return true;
      }
      string input = ((JsonElement) ref element).GetString();
      Match match = this._regex.Match(input);
      if (match.Value != input)
      {
        errorMessage = "Value is not a match for the format '" + this.Key + "'.";
        return false;
      }
      errorMessage = (string) null;
      return match.Value == input;
    }
  }
}
```

Also, the special regex highlighting of Resharper does not work.
image

Again, I am just trying to understand the use of the annotations here. I really love your work and other than this one issue the nuget packages are a pleasure to use :-)

gregsdennis:
🤔 I definitely intend that support to be available.  I'll do some digging.  It is my feeling that this is a required dependency.

gregsdennis:
I've added the build symbol to the project and the attribute now appears in the decompilation.  Thanks for the tip.  It'll go out with the next version.  I don't expect that'll be too long.

PR #296, opened 2022-07-03
Check Condition on System.Text.Json.JsonIgnoreAttribute

ivarne:
I'm typically adding
`[JsonIgnore(Condition = JsonIgnoreCondition.Never)]`
to my classes, to ensure that serialized json has the format I want,
even if someone changes the global `JsonIgnoreCondition`.

This means that the existance of `JsonIgnore` is not enough, but you
need to check that `Condition == JsonIgnoreCondition.Always`.

The Docs for `[JsonIgnore]` states that `Always` is the default condition
and the other conditions does not indicate that the property should be
ignored from schema.
https://docs.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonignoreattribute.condition?view=net-6.0

### DIFF
--- a/JsonSchema.Generation/Generators/ObjectSchemaGenerator.cs
+++ b/JsonSchema.Generation/Generators/ObjectSchemaGenerator.cs
@@ -43,7 +43,7 @@ internal class ObjectSchemaGenerator : ISchemaGenerator
 			var memberAttributes = member.GetCustomAttributes().ToList();
 #pragma warning disable 8600 // Assigning null to non-null
 			// ReSharper disable once AssignNullToNotNullAttribute
-			var ignoreAttribute = (Attribute)memberAttributes.OfType<JsonIgnoreAttribute>().FirstOrDefault() ??
+			var ignoreAttribute = (Attribute)memberAttributes.OfType<JsonIgnoreAttribute>().Where(a=>a.Condition == JsonIgnoreCondition.Always).FirstOrDefault() ??
 								  memberAttributes.OfType<JsonExcludeAttribute>().FirstOrDefault();
 #pragma warning restore 8600
 			if (ignoreAttribute != null) continue;


JsonSchema.Generation/Generators/ObjectSchemaGenerator.cs
```diff
@@ -43,7 +43,7 @@ public void AddConstraints(SchemaGenerationContextBase context)
 			var memberAttributes = member.GetCustomAttributes().ToList();
 #pragma warning disable 8600 // Assigning null to non-null
 			// ReSharper disable once AssignNullToNotNullAttribute
-			var ignoreAttribute = (Attribute)memberAttributes.OfType<JsonIgnoreAttribute>().FirstOrDefault() ??
+			var ignoreAttribute = (Attribute)memberAttributes.OfType<JsonIgnoreAttribute>().Where(a=>a.Condition == JsonIgnoreCondition.Always).FirstOrDefault() ??
```
gregsdennis, line 46: I fear this will hide the `[JsonIgnore]` attribute if the ignore condition isn't `Always`.  Would it be better to get the attribute if it exists at all, _then_ check the condition?

gregsdennis (commented):
Thanks for catching this, @ivarne.

JsonSchema.Generation/Generators/ObjectSchemaGenerator.cs
gregsdennis: Right, but I'm concerned doing this in a `.Where()` will _completely_ skip that the attribute is there, which is _not_ what we want if one of the other values is used.  I think it's better to get the attribute, then have the check for its value in a separate `if` statement.

ivarne: I’m really sorry, but I don’t understand what you are requesting in terms of behavior change.

The old code check for the existence of `System.Text.Json.[JsonIgnore]` or your custom `[JsonExclude]`, (in a very compact way) and ignores the whole property with `continue;` on the next line. The individual attributes are never read.

gregsdennis (approved):
I've changed the target branch to a version branch so I can edit version and release notes, but this looks fine.

Thanks!

JsonSchema.Generation/Generators/ObjectSchemaGenerator.cs
ivarne: Aha, so you were talking about the what would happen if both `JsonIgnore` and `JsonExclude`, would be set? Sorry for not understanding.

Should probably include some tests for all the possible combinations of the attributes, but wanted to clarify intended behavior first.

PR #303, opened 2022-07-16
make rules public (ctors still internal)

### LINKED ISSUE #286: An exception is thrown if the "var" operator returns null from some operators
**Environment**
- Nuget Package: JsonLogic 2.0.0
- Nuget Version: 6.1.0
- OS: WSL
- .Net Target: net6.0

**Describe the bug**
An exception is thrown if the "var" operator returns null from some operators. 
image

**To Reproduce**
```
[Test]
public void ExceptionThrownIfVarNull ()
{
  var rule = System.Text.Json.JsonSerializer.Deserialize<Rule>("{ \"in\" : [ {\"var\": \"item\"}, {\"var\": \"list\"} ] }");
  var result = rule.Apply();
}
```

**Expected behavior**
rule.Apply() should return false if "var" operator returns null.
if data was not been provided.
image
if var property could not be found.
image
for JsonLogic "play  ground"
```json
{"in": [{"var": "item"}, {"var": "list"}]}
```

**Additional context**
This behavior could be found only in some methods. I only know about "in" and "<" but I think it could be some more.
For method like "===" is working as expected
image
```c#
[Test]
public void JustExample ()
{
	var rule = System.Text.Json.JsonSerializer.Deserialize<Rule>("{ \"===\" : [ {\"var\": \"item\"}, {\"var\": \"list\"} ] }");
	var result = rule.Apply();
}
```
Not sure if null === null is correct. Thought.

For "in" operator it should be a simple fix:
```c#
public override JsonNode? Apply(JsonNode? data, JsonNode? contextData = null)
{
	var test = _test.Apply(data, contextData);
	var source = _source.Apply(data, contextData);

	if (source is JsonValue value && value.TryGetValue(out string? stringSource))
	{
		var stringTest = test.Stringify();

		if (stringTest == null || stringSource == null)
			throw new JsonLogicException($"Cannot check string for {test.JsonType()}.");

		return !string.IsNullOrEmpty(stringTest) && stringSource.Contains(stringTest);
	}

	// suggestion
	if (source == null)
		return false;

	if (source is JsonArray arr)
		return arr.Any(i => i.IsEquivalentTo(test));

	throw new JsonLogicException($"Cannot apply `in` to {test.JsonType()} and {source.JsonType()}.");
}
```

Not sure about other rules.

Please could you apply a solution to both JsonLogic 2.0.0 and 3.00

### LINKED ISSUE #302: Operator return type
**Environment**
- Nuget Package: JsonLogic 3.0.1
- Nuget Version: 6.1.0
- OS: WSL
- .Net Target: Net6.0

**How can I help you today?**
Hello, is there any way to validate if the operator returns a boolean as result?

**Additional context**
I need to find a way to check if JsonLogic expression is valid for the custom cases. In my case, only boolean expressions are allowed. Rules classes are internal, so I could not use them to check the type after Deserialization. It would be good if rule class has a return type property or some validator that consumes some custom rules.
Currently, the only way to validate if an expression is correct is to convert it?

gregsdennis:
Resolves #302 - I'm not going to put a return type property on the rule, but it's likely something you can make an extension method for pretty easily.

Resolves #286 (attempted anyway) - added a cast from null to 0.  Hopefully that resolves a bunch of what was going on there.

### DIFF
--- a/Json.More/Json.More.csproj
+++ b/Json.More/Json.More.csproj
@@ -17,14 +17,15 @@
     <LangVersion>latest</LangVersion>
     <DocumentationFile>Json.More.xml</DocumentationFile>
     <PackageIcon>json-logo-256.png</PackageIcon>
-	<IncludeSymbols>true</IncludeSymbols>
-	<SymbolPackageFormat>snupkg</SymbolPackageFormat>
+		<IncludeSymbols>true</IncludeSymbols>
+		<SymbolPackageFormat>snupkg</SymbolPackageFormat>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>../json-everything.snk</AssemblyOriginatorKeyFile>
-	<Nullable>enable</Nullable>
-   </PropertyGroup>
+	  <NoWarn>CS1570</NoWarn>
+		<Nullable>enable</Nullable>
+  </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="System.Text.Json" Version="6.0.2" />
--- a/JsonLogic.Tests/AddTests.cs
+++ b/JsonLogic.Tests/AddTests.cs
@@ -84,10 +84,10 @@ public class AddTests
 	}
 
 	[Test]
-	public void AddSingleNullThrowsError()
+	public void AddSingleNullReturns0()
 	{
 		var rule = new AddRule(LiteralRule.Null);
 
-		Assert.Throws<JsonLogicException>(() => rule.Apply());
+		JsonAssert.AreEquivalent(0, rule.Apply());
 	}
 }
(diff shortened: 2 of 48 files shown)


gregsdennis:
It doesn't look like I can make the constructors internal because of serialization:

```
Constructor on type 'Json.Logic.Rules.VariableRule' not found.
```

I definitely don't want to expose those.  I want people to use the `JsonLogic` builder methods.

jevvo-trimble:
Please try this: 
LogicComponentConverter.Rule line 100
```
return (Rule)Activator.CreateInstance(ruleType,
    BindingFlags.NonPublic | BindingFlags.Instance,
    null,
    value.Cast<object>()
	    .Select(o => o ?? new LiteralRule(null))
	    .ToArray(),
    null);
```
Test for it:
```
[Test]
public void Test()
{
    var rule = JsonSerializer.Deserialize<Rule>("{\"var\": \"data\"}");
}
```
https://stackoverflow.com/questions/4077253/instantiating-a-constructor-with-parameters-in-an-internal-class-with-reflection#:~:text=The%20issue%20is,typeToInstantiate%2C%20true)

gregsdennis:
The problem isn't that the _types_ are internal; it's that the _constructors_ are.  I have an idea...

gregsdennis:
okay.  I had to add converters for all the rules as well as tweak the deserialization process, but it seems to work now.

PR #305, opened 2022-07-17
Minor CSS fix

drewdelano:


gregsdennis:
Thanks.  This led me to a good solution overall.

PR #311, opened 2022-08-02
Add test cases for ignoring properties

ivarne:
The discussion in https://github.com/gregsdennis/json-everything/pull/296 seems to indicate that there should be some tests to ensure that the ignore attributes are read in correct order.

* `[JsonExclude]` will override for schema generation, regardless of other properties present.
* `[JsonIgnore]` will only be used to ignore properties when the condition is `Condition= JsonIgnoreCondition.Always`, `[JsonIgnore(Condition= JsonIgnoreCondition.Never)]`, will not cause the property to be removed.


gregsdennis:
Nice.  Thanks for these!

PR #316, opened 2022-08-12
Vocab data v2

gregsdennis:
Resolves some conversations with @handrews in the JSON Schema Slack workspace.

Change summary:

- instances are now referenced using either plain or relative JSON Pointers
- schema data is now referenced using JSON Pointer fragments without needing to specify the full URI

This brings `data` resolution in closer alignment with `$ref` (and potentially solves some other issues.

### DIFF
--- a/json-everything.net/wwwroot/md/vocabs-data-2022.md
+++ b/json-everything.net/wwwroot/md/vocabs-data-2022.md
@@ -36,8 +36,8 @@ The value of `data` must be an object.  The keys of the object are interpreted a
 
 - JSON Pointers per [RFC 6901](https://www.rfc-editor.org/rfc/rfc6901) 
 - [Relative JSON Pointers](https://json-schema.org/draft/2019-09/relative-json-pointer.html)
-- URI-encoded JSON Pointer identifier per [RFC 6901, §6](https://www.rfc-editor.org/rfc/rfc6901#section-6)
-- Absolute URIs per [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986), optionally with a JSON Pointer fragment identifier
+- fragment-only IRI using IRI-encoded JSON Pointer identifier per [RFC 6901, §6](https://www.rfc-editor.org/rfc/rfc6901#section-6)
+- Absolute IRIs per [RFC 3987](https://datatracker.ietf.org/doc/html/rfc3987), optionally including an additional IRI-encoded JSON Pointer fragment identifier
 
 `data` MUST NOT contain any keys which are defined by the JSON Schema Core Vocabulary.
 
@@ -52,12 +52,14 @@ The validation and annotation results of `data` are those of the formed schema. 
 
 `data` MUST be processed contextually in the same manner as the host schema.  Specifically,
 
-- URI resolution MUST be performed in the same manner as `$ref` (per section 3.3).
+- IRI resolution MUST be performed in the same manner as `$ref` (per section 3.3).
 - Keys that are ignored by the host schema MUST also be ignored by the formed schema.
 
-Implementations SHOULD validate the formed schema against the host schema's meta-schema (as specified by `$schema`) to ensure that it is a syntactically valid schema object.
+Implementations SHOULD validate that the resolved data forms a valid schema against the host schema's meta-schema (as specified by `$schema`).
 
-### 4.3 URI Resolution
+***NOTE** It is not necessary for an implementation to validate using the meta-schema.  Other mechanisms internal to the implementation (such as deserialization) may suffice to perform this task.*
+
+### 4.3 IRI Resolution
 
 The values of `data` are dereferenced in different ways depending on the format of the value.
 
@@ -67,7 +69,7 @@ If the value is a Relative JSON Pointer, it is resolved against the instance at 
 
 Otherwise, it must be resolved in accordance with the rules of `$ref` resolution for the relevant JSON Schema specification (e.g. [draft 2020-12, §8.2](https://json-schema.org/draft/2020-12/json-schema-core.html#name-base-uri-anchors-and-derefe)).  However, unlike `$ref` which requires that the indicated data must represent a valid schema object, a `data` reference can identify any value which is valid for the associated keyword.
 
-Because JSON Pointers and Relative JSON Pointers are syntactically valid URIs, the value MUST be checked in the sequence indicated above in order to properly identify the pointer types from other URIs.
+Because JSON Pointers and Relative JSON Pointers are syntactically valid IRIs, the value MUST be checked in the sequence indicated above in order to properly identify the pointer types from other IRIs.
 
 For each successfully resolved reference, the full value at the specified location MUST be returned.
 
@@ -75,7 +77,7 @@ If a reference cannot be resolved, or if a resolved value is not valid for the a
 
 #### 4.3.1 External Data Access
 
-Implementations SHOULD provide a means to pre-load and cache any external references prior to evaluation but MAY be configured to fetch external documents at evaluation time.  Documents fetched from URIs which contain a JSON Pointer fragment MUST be interpreted using a media type that allows resolution of such fragments.
+Implementations SHOULD provide a means to pre-load and cache any external references prior to evaluation but MAY be configured to fetch external documents at evaluation time.  Documents fetched from IRIs which contain a JSON Pointer fragment MUST be interpreted using a media type, such as `application/schema-instance+json`, that allows resolution of such fragments.
 
 Users should be aware that fetching data from external locations may carry certain security risks not covered by this document.
 
--- a/json-everything.net/wwwroot/md/vocabs-data-2022.md
+++ b/json-everything.net/wwwroot/md/vocabs-data-2022.md
@@ -77,7 +77,7 @@ If a reference cannot be resolved, or if a resolved value is not valid for the a
 
 Implementations SHOULD provide a means to pre-load and cache any external references prior to evaluation but MAY be configured to fetch external documents at evaluation time.  Documents fetched from URIs which contain a JSON Pointer fragment MUST be interpreted using a media type that allows resolution of such fragments.
 
-Users should be aware that fetching data from external location may carry certain security risks not covered by this document.
+Users should be aware that fetching data from external locations may carry certain security risks not covered by this document.
 
 ### 4.4 Output
 
--- a/json-everything.net/wwwroot/md/vocabs-data-2022.md
+++ b/json-everything.net/wwwroot/md/vocabs-data-2022.md
@@ -1,0 +1,123 @@
+# A Vocabulary for Accessing Data Stored in JSON
+
+## 1. Purpose
+
+This document describes a vocabulary defining keywords that can be used to reference values stored in
+
+- the instance data
+- external JSON data
+- the schema data
+
+and use the dereferenced values as input for other keywords.
+
+The intent for this keyword is to cover the use cases discussed across several issues in the JSON Schema specification GitHub repositories.  (A quick search for `"$data"` can readily summon these issues.)
+
+## 2. Declarations
+
+The ID for this vocabulary is `https://json-everything.net/vocabs-data-2022` (the URI to this document).
+
+A draft 2020-12 meta-schema which includes this vocabulary has been defined for convenience.  The ID for the meta-schema is `https://json-everything.net/meta/data-2022`, and it can also be found at this address.
+
+## 3. The `data` Keyword
+
+### 3.1 Syntax and Semantics
+
+The value of `data` must be an object.  The keys of the object correspond to valid JSON Schema keywords, and the values MUST be one of
+
+- JSON Pointers per [RFC 6901](https://www.rfc-editor.org/rfc/rfc6901)
+- [Relative JSON Pointers](https://json-schema.org/draft/2019-09/relative-json-pointer.html)
+- URI references per [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986), potentially with JSON Pointer fragments
+
+The set of keywords in the schema or subschema containing `data` SHOULD be distinct from the set of keys defined within `data`.  Behavior in the event of an overlap is not defined.
+
+`data` operates in two phases:
+
+1. All of the values are dereferenced per sections 3.2 and 3.3.
+2. The keywords and values then form a schema which is applied to the instance at the current location.
+
+The validation and annotation results of `data` are those of the formed schema.  More detail regarding output can be found in section 3.4.
+
+### 3.2 Contextual Behavior
+
+`data` MUST be processed contextually in accordance with the draft of the schema in which it is used.  For example, if `data` is used in a schema that declares draft 2019-09, then
+
+- the keys SHOULD be recognized by draft 2019-09 or defined in other referenced vocabularies, and
+- the URI references MUST be resolved following the same rules as `$ref` as defined by draft 2019-09.
+
+If a key is not a recognized keyword in the schema's draft (e.g. `$dynamicAnchor` in a draft 2019-09 schema), then this key is ignored just as it would be ignored if it were in the schema or subschema containing `data`.
+
+### 3.3 URI Resolution
+
+The values of `data` are dereferenced in different ways depending on the format of the value.
+
+If the value is a JSON Pointer, it is resolved against the instance root.
+
+If the value is a Relative JSON Pointer, it is resolved against the instance at the location currently being evaluated.
+
+If the value is a URI reference, it must be resolved in accordance with the rules of `$ref` resolution for the containing spec, as per section 3.2.  The primary different between resolving URIs for `data` and resolving URIs for `$ref` is that `$ref` requires that the indicated data must represent a valid schema object whereas `data` can reference any valid JSON value.
+
+External document data can be referenced using an absolute URI base identifier along with a JSON Pointer fragment.
+
+Implementations SHOULD be able to download external JSON documents, but MAY provide means to pre-load and cache such documents against the identifiers by which they will be referenced.  Documents downloaded from URIs which contain a JSON Pointer fragment MUST be interpreted using a media type that allows resolution of such fragments.
+
+If any single reference cannot be resolved, validation MUST fail; otherwise the full value at the specified location is returned.
+
+If the resolved value is not valid for the associated keyword, validation MUST fail.
+
+### 3.4 Errors
+
+The output formatting specified by the JSON Schema Core specification can only indicate that something failed at the `data` node, but there is no provision for providing further detail.
+
+In summary, the `data` keyword can generate two validation failure states as specified by this document:
+
+- a URI reference cannot be resolved, and
+- a URI reference can be resolved but the returned value is invalid for the associated keyword.
+
+To make debugging `data` simpler, implementations SHOULD provide an error message indicating what the failure was and for which key it occurred.
+
+If both of these succeed, the validation output of the resulting subschema is reported into the overall schema output as if the subschema were actually a child of the `data` keyword.
+
+## 4. A Short Example
+
+The following defines a schema to validate an object instance with a `foo` property that must contain an integer value less than or equal to the value in the instance's `minValue` property.
+
+```json
+{
+  "$schema": "https://json-everything.net/meta/data-2022",
+  "type": "object",
+  "properties": {
+    "foo": {
+      "type": "integer",
+      "data": {
+        "minimum": "/minValue"
+      }
+    },
+    "minValue": {
+      "type": "integer"
+    }
+  },
+  "dependentRequired": {
+    "foo": [ "minValue" ]
+  }
+}
+```
+
+The `data` property declares that its parent subschema should validate a `minimum` keyword whose value is the value in the `minValue` property of the instance.
+
+Note also the use of `dependentRequired` to ensure that when `foo` is present in the instance, `minValue` is also present.  While this not explicitly required of schema authors, this type of insurance mechanism is recommended.
+
+The following shows how a change in the `minValue` property can affect the validation result of the `foo` property and thus the entire instance.
+
+```json
+// passing
+{
+  "minValue": 5,
+  "foo": 10
+}
+
+// failing
+{
+  "minValue": 15,
+  "foo": 10
+}
+```
--- a/json-everything.net/wwwroot/md/vocabs-data-2022.md
+++ b/json-everything.net/wwwroot/md/vocabs-data-2022.md
@@ -5,10 +5,10 @@
 This document describes a vocabulary defining keywords that can be used to reference values stored in
 
 - the instance data
-- external JSON data
 - the schema data
+- external JSON data
 
-and use the dereferenced values as input for other keywords.
+where the dereferenced values serve as input for keywords in a derived subschema.
 
 The intent for this keyword is to cover the use cases discussed across several issues in the JSON Schema specification GitHub repositories.  (A quick search for `"$data"` can readily summon these issues.)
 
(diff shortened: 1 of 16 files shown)


json-everything.net/wwwroot/md/vocabs-data-2022.md
```diff
@@ -0,0 +1,123 @@
...
+
+If the value is a JSON Pointer, it is resolved against the instance root.
+
+If the value is a Relative JSON Pointer, it is resolved against the instance at the location currently being evaluated.
+
+If the value is a URI reference, it must be resolved in accordance with the rules of `$ref` resolution for the containing spec, as per section 3.2.  The primary different between resolving URIs for `data` and resolving URIs for `$ref` is that `$ref` requires that the indicated data must represent a valid schema object whereas `data` can reference any valid JSON value.
+
+External document data can be referenced using an absolute URI base identifier along with a JSON Pointer fragment.
+
+Implementations SHOULD be able to download external JSON documents, but MAY provide means to pre-load and cache such documents against the identifiers by which they will be referenced.  Documents downloaded from URIs which contain a JSON Pointer fragment MUST be interpreted using a media type that allows resolution of such fragments.
```
handrews, line 61: I forgot something:  It would better align with JSON Schema's policies on remote resources in general, and `$ref` resolution in particular, if you switched this so that the pre-load cache is a SHOULD while downloading external documents is a MAY.

It's also not realistic to expect an implementation to support all possible URL schemes for downloads, plus URN-like URI schemes can't be downloaded anyway.

json-everything.net/wwwroot/meta/data-2022
```diff
@@ -0,0 +1,31 @@
+{
+    "$id": "https://json-everything.net/meta/data-2022",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$vocabulary": {
+        "https://json-schema.org/draft/2020-12/vocab/core": true,
+        "https://json-schema.org/draft/2020-12/vocab/applicator": true,
+        "https://json-schema.org/draft/2020-12/vocab/validation": true,
+        "https://json-schema.org/draft/2020-12/vocab/meta-data": true,
+        "https://json-schema.org/draft/2020-12/vocab/format-annotation": true,
```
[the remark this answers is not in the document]
gregsdennis: Oversight & carryover from previous version.

json-everything.net/wwwroot/md/vocabs-data-2022.md
```diff
@@ -0,0 +1,123 @@
...
+
+### 3.1 Syntax and Semantics
+
+The value of `data` must be an object.  The keys of the object correspond to valid JSON Schema keywords, and the values MUST be one of
+
+- JSON Pointers per [RFC 6901](https://www.rfc-editor.org/rfc/rfc6901)
+- [Relative JSON Pointers](https://json-schema.org/draft/2019-09/relative-json-pointer.html)
+- URI references per [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986), potentially with JSON Pointer fragments
+
+The set of keywords in the schema or subschema containing `data` SHOULD be distinct from the set of keys defined within `data`.  Behavior in the event of an overlap is not defined.
```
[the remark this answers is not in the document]
gregsdennis: The original intent was that rather than `data` operating as an applicator (as we've discussed), the result would be "merged" with the parent schema.  That would mean that keyword conflicts could arise.

I think `data` operating as an applicator resolves this, so, yeah, this could just come out.

json-everything.net/wwwroot/md/vocabs-data-2022.md
```diff
@@ -0,0 +1,123 @@
...
+The validation and annotation results of `data` are those of the formed schema.  More detail regarding output can be found in section 3.4.
+
+### 3.2 Contextual Behavior
+
+`data` MUST be processed contextually in accordance with the draft of the schema in which it is used.  For example, if `data` is used in a schema that declares draft 2019-09, then
+
+- the keys SHOULD be recognized by draft 2019-09 or defined in other referenced vocabularies, and
+- the URI references MUST be resolved following the same rules as `$ref` as defined by draft 2019-09.
+
+If a key is not a recognized keyword in the schema's draft (e.g. `$dynamicAnchor` in a draft 2019-09 schema), then this key is ignored just as it would be ignored if it were in the schema or subschema containing `data`.
```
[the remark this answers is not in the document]
gregsdennis: Again, redefining `data` as an applicator resolves a lot of this.

json-everything.net/wwwroot/md/vocabs-data-2022.md
```diff
@@ -0,0 +1,123 @@
...
+
+The values of `data` are dereferenced in different ways depending on the format of the value.
+
+If the value is a JSON Pointer, it is resolved against the instance root.
+
+If the value is a Relative JSON Pointer, it is resolved against the instance at the location currently being evaluated.
+
+If the value is a URI reference, it must be resolved in accordance with the rules of `$ref` resolution for the containing spec, as per section 3.2.  The primary different between resolving URIs for `data` and resolving URIs for `$ref` is that `$ref` requires that the indicated data must represent a valid schema object whereas `data` can reference any valid JSON value.
+
+External document data can be referenced using an absolute URI base identifier along with a JSON Pointer fragment.
```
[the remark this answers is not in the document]
gregsdennis: > Is there any real need to require an absolute URI?

If resolution is handled as it is for `$ref`, then I think I can remove this.  I just need a cleaner way of stating that.

json-everything.net/wwwroot/md/vocabs-data-2022.md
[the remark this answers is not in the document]
gregsdennis: > So you definitely want to restrict this to JSON documents?

Not necessarily.  You're right in that it needs to be mappable to JSON.  However, I don't know that I can realistically support non-JSON, much less "any mappable format."  That's really open-ended and not practical.

json-everything.net/wwwroot/md/vocabs-data-2022.md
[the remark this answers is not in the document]
handrews: Awesome - yeah it's interesting that neither of us looked at `data` as an applicator at first (I only called it "kind of like an applicator"), but once we agreed on that way of looking at it things got simpler all around.  That's really the kind of thing I've always hoped the classifications would help with.

json-everything.net/wwwroot/md/vocabs-data-2022.md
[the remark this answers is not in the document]
handrews: I think all you need to say is that base URI for URI-references in `data` values is determined per [§8.2 of the core specification](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#name-base-uri-anchors-and-derefe).  That's what controls `$ref`'s behavior, so at that point you don't even need the explicit comparison to `$ref` (although you could note that this is the same process if you think that makes for clearer documentation).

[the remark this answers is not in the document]
handrews: Or if you don't want to reference a single spec version, just say "according to the base URI and URI-reference resolution rules of the appropriate draft" and maybe point to the 2019-09 and 2020-12 sections so that people know exactly what you mean.

json-everything.net/wwwroot/md/vocabs-data-2022.md
```diff
@@ -0,0 +1,123 @@
...
+
+Implementations SHOULD be able to download external JSON documents, but MAY provide means to pre-load and cache such documents against the identifiers by which they will be referenced.  Documents downloaded from URIs which contain a JSON Pointer fragment MUST be interpreted using a media type that allows resolution of such fragments.
+
+If any single reference cannot be resolved, validation MUST fail; otherwise the full value at the specified location is returned.
+
+If the resolved value is not valid for the associated keyword, validation MUST fail.
```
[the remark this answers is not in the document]
gregsdennis: > Which could be from an unknown vocabulary if this is combined with other extension vocabularies.

I may need to add a restriction that core-vocab keywords (`$*`) are explicitly disallowed.  I'm not sure what use having `$id` or `$schema` pulled from an external location would be.

Adding this restriction would have `data` use the parent schema's meta-schema and would implicitly require that any keywords are from the vocabs already in use.

json-everything.net/wwwroot/md/vocabs-data-2022.md
```diff
@@ -0,0 +1,123 @@
...
+
+### 3.4 Errors
+
+The output formatting specified by the JSON Schema Core specification can only indicate that something failed at the `data` node, but there is no provision for providing further detail.
```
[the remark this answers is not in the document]
gregsdennis: Yep.  Again... `data` as an applicator.

json-everything.net/wwwroot/md/vocabs-data-2022.md
gregsdennis: I went with this primarily because a lot of the original discussion was around fetching data from external sources, possibly unknown until evaluation time.  Aligning with the spec is a good argument, but I do want to leave room for fetching data.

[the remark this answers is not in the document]
handrews: This is why I was pushing for an emphasis on pre-populating a URI-to-data storage (like the schema storage), because if you want to go outside of HTTP and JSON, that can be done by an app that fetches, parses, and maps the data into the data model, and then `data` just looks it up by URI.

If `data` has to do the fetching, then we're pretty much limited to a small set of URI schemes/protocols and media types.

If the pre-load behavior becomes the recommended default, and the auto-fetch an optional extra, then limiting the auto-fetch to HTTP(S) and JSON-compatible media types is fine, as it doesn't constrain the overall usage.

I'm particularly thinking of a usage of `data` to access values from a database that is not served over HTTP.

json-everything.net/wwwroot/md/vocabs-data-2022.md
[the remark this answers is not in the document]
handrews: Ah, yeah, that is probably a good idea.  You don't want the template subschema to be something you can reference by URI, as it's not meaningful: only a specific resolved version would be a meaningful reference target.

So `$id` and `$schema` should be out.  `$ref` and `$dynamicRef` are problematic because a URI-reference value would be ambiguous (is it the `$ref` target or a `data` template location that provides the `$ref` target?), and `$ref`-ing some runtime data is like `eval()`-ing a random string of JavaScript code.  Definitely a security concern.

json-everything.net/wwwroot/md/vocabs-data-2022.md
handrews: I think the only distinction here is whether the application fetches the data immediately beforehand and passes it to the implementation (somehow... we don't have a notion of populating a non-schema data store as part of the implementation interface), vs actually re-doing the fetch every time the `data` keyword is hit.

How important is it to re-do the fetch every time?

Anyway, I'm not dead-set against fetching, it just feels a little outside of JSON Schema norms.  Then again, I'm not sure how a vocabulary can require a pre-populatable data cache, so... 🤔

json-everything.net/wwwroot/md/vocabs-data-2022.md
```diff
@@ -0,0 +1,121 @@
...
+
+### 3.3 URI Resolution
+
+The values of `data` are dereferenced in different ways depending on the format of the value.
+
+If the value is a JSON Pointer, it is resolved against the instance root.
+
+If the value is a Relative JSON Pointer, it is resolved against the instance at the location currently being evaluated.
+
+If the value is a URI reference, it must be resolved in accordance with the rules of `$ref` resolution for the relevant JSON Schema specification (e.g. [draft 2020-12, §8.2](https://json-schema.org/draft/2020-12/json-schema-core.html#name-base-uri-anchors-and-derefe)).  However, unlike `$ref` which requires that the indicated data must represent a valid schema object, a `data` reference can identify any value which is valid for the associated keyword.
```
handrews, line 57: I just realized a significant problem, and you should totally laugh at me for this after all of my pontificating about understanding URI-references properly...

Both JSON Pointers and Relative JSON Pointers are syntactically valid URI-references (treated as paths rather than fragments — so the effect would be completely different), which means there is no way to automatically detect which is which.

I think your options are:

1. split this into two keywords (and given the relative difficulty and security concerns of instance vs external support, probably separate vocabularies as the instance one is far more likely to be implemented by others), using the schema's base URI for the external keyword
2. go back to mandating full URIs (starting with a scheme) for all non-JP/RJP values.  (Note that these are not absolute-URIs, which start with a scheme but cannot have a fragment).

Option 2 would preserve the UX you want to have, and as far as accessing schema data, would be no worse than the current situation.  I know you're against option 1, I'm just including it for completeness.

gregsdennis: > Both JSON Pointers and Relative JSON Pointers are syntactically valid URI-references

```
/is/this/a/uri/or/a/pointer
```

🤔 yeah, I kinda see that.

I much prefer the single keyword.  Having all of the references in one place, creating a single subschema, just makes sense from a usability standpoint.  It's probably a bit harder on an implementer, but as a schema author, I'd hate to have to manage this in two places.

I see a few cases that I'd want to cater to here.

```
/plain/json/pointer
2/relative/json/pointer
#/pointer/to/a/location/in/the/schema      (uri-encoded pointer)
#anchor-within-the-schema                  (anchor syntax)
https://explicit.ly/external/data          (absolute uri)
```

I think that each of these should be uniquely identifiable.

handrews: > Having all of the references in one place, creating a single subschema, just makes sense from a usability standpoint.

While I see your argument, I really don't agree.  To me it's a lot easier to have two things that each have a very clear purpose and syntax.  No chance of confusing a pointer and a relative path.  Not to mention, far more likely to find support for the easier vocabulary to implement.  I don't really understand at all why combining two schema templates is problematic - there are tons of reasons we combine multiple schemas, it's a normal thing to do.
  
> I think that each of these should be uniquely identifiable.

Then you have to introduce limits on what URIs are valid, which is another headache.  I find that plus the need to duplicate the schema `$id` if you want to reference it to be bigger problems than having to use two schema templates side by side (and that's as a schema author, not a plugin writer).

But again, this is your proposed vocabulary, I'm explaining myself, not demanding that you do what I'd prefer.

gregsdennis: 🤦 `#anchor-within-the-schema` isn't necessary to support.  It's only realistic inside a schema object, at which point, you'd just use `$ref` anyway.

---

> plus the need to duplicate the schema `$id` if you want to reference it

That's why I included `#/pointer/to/a/location/in/the/schema` as an option.  This allows you to get to values within the same schema.  If you want to reference an external schema, yeah, you'd need to use the full, absolute `$id` URI for it.

---

Just for fun, let's look at an example of multiple keywords.

```json
{
  "$schema": "https://gregsdennis.github.io/json-everything/meta/data",
  "$id": "https://examp.le/schema",
  "type": "object",
  "properties": {
    "foo": {
      "type": "integer",
      "data": {
        "minimum": "/minValue",
        "maximum": "https://examp.le/data#/maxValue",
        "multipleOf": "#/values/multiples"
      }
    },
    "minValue": {
      "type": "integer"
    }
  },
  "values": {
    "multiples": 5
  },
  "dependentRequired": {
    "foo": [ "minValue" ]
  }
}

{
  "$schema": "https://gregsdennis.github.io/json-everything/meta/data",
  "$id": "https://examp.le/schema",
  "type": "object",
  "properties": {
    "foo": {
      "type": "integer",
      "instanceData": {
        "minimum": "/minValue"
      },
      "externalData": {
        "maximum": "data#/maxValue",
        "multipleOf": "#/values/multiples"
      }
    },
    "minValue": {
      "type": "integer"
    }
  },
  "values": {
    "multiples": 5
  },
  "dependentRequired": {
    "foo": [ "minValue" ]
  }
}
```

In the first example,

- we dereference all of the locations which forms a single schema object, then we process that schema
- all of the data references are under a single keyword, meaning the author only has to look in one place for them  

In the second example,

- we dereference the locations under `instanceData` which forms a schema object, then we process that schema.  Then we dereference the locations under `externalData` which also forms a schema object, and we process that one.
- the data references are split between two keywords, meaning the author needs to be sure to look in both
- is `#/values/multiples really "external" or would we need a third keyword for schema locations (`internalData`)?

***NOTE** Yes, I realize that my "internal" example isn't great, but it gets the point across.  I expect an author would want to pull a value from some other keyword elsewhere in the schema, not just a miscellaneous value store.*

Yes, the syntax of the external reference URIs are simpler in the two-keyword case, but I don't think that benefit outweighs the convenience and readability in the first.

handrews: > That's why I included #/pointer/to/a/location/in/the/schema as an option. This allows you to get to values within the same schema. If you want to reference an external schema, yeah, you'd need to use the full, absolute $id URI for it.

Oh, right, idk why I missed that.

> the data references are split between two keywords, meaning the author needs to be sure to look in both

I'm really not trying to be difficult, I just don't understand why that's a concern.  Clearly it is for you!  But it really makes no difference at all to me.  Although I guess if you needed to mix data sources on keywords that read adjacent annotations only... hmm... that is mostly for applicators, though, which would involve... loading a schema from within the instance (which is another security concern)?  I can't imagine why you'd use `data` for an external schema when you could just use `$ref`.

> is `#/values/multiples` really "external" or would we need a third keyword for schema locations (internalData).

It's external to the instance, but I agree that the terminology here is not great.  I'm using "external" to mean "referenced by URI", pretty much, and I realize that comes across weirdly with schema data.

> Yes, the syntax of the external reference URIs are simpler in the two-keyword case, but I don't think that benefit outweighs the convenience and readability in the first.

It doesn't.  Really, the most serious objection to me is the fact that people (which perhaps just means "me") aren't as likely to implement something that requires external data access.  But that only matters if maximum adoption is your goal.  If you're more focused on implementing the use case than seeing it implemented broadly (which is totally valid), then that concern doesn't matter much.

gregsdennis: > people ... aren't as likely to implement something that requires external data access

Are you suggesting that `instanceData` and `externalData` (or whatever) are created as two separate vocabs rather than two keywords in the same vocab so that they can be implemented separately?

handrews: @gregsdennis 
> Are you suggesting that `instanceData` and `externalData` (or whatever) are created as two separate vocabs rather than two keywords in the same vocab so that they can be implemented separately?

Yes.  Normally I'd prefer grouping related functionality, but one thing we learned from `format` is that if you combine simple things with difficult things, people either don't implement it at all or half-ass the whole thing.

I'd probably make two vocabularies:
* one with `instanceData` (JP & RJP) and `schemaData` (URIs/IRIs with the current schema location's base URI/IRI for resolving relative URI/IRI-references, but it resolves the value from the schema storage just like `$ref`)
* one that's just `externalData` (URIs/IRIs, but it resolves them from some separate source, either by fetching or a data storage of some sort)

gregsdennis: I hear you, but I'm not prepared to make multiple vocabs for data reference functionality at this time.  I think what I have (plus some pending edits) to refine the single `data` keyword is the proper next evolution for this vocab.

gregsdennis:
With the latest edits, I have a few more changes to make to the implementation.

- [ ] Add a data registry
- [ ] Update the data fetch to _not_ fetch by default
- [ ] Still need to fix (and probably expand) the tests
- [ ] Maybe need an interoperable test suite? (might be good for a separate PR)

json-everything.net/wwwroot/md/vocabs-data-2022.md
```diff
@@ -16,66 +16,76 @@ The intent for this keyword is to cover the use cases discussed across several i
...
+
+### 4.1 Syntax and Semantics
 
 The value of `data` must be an object.  The keys of the object are interpreted as JSON Schema keywords, and the values MUST be one of
 
-- JSON Pointers per [RFC 6901](https://www.rfc-editor.org/rfc/rfc6901)
+- JSON Pointers per [RFC 6901](https://www.rfc-editor.org/rfc/rfc6901) 
 - [Relative JSON Pointers](https://json-schema.org/draft/2019-09/relative-json-pointer.html)
-- URI references per [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986), potentially with JSON Pointer fragments
+- URI-encoded JSON Pointer identifier per [RFC 6901, §6](https://www.rfc-editor.org/rfc/rfc6901#section-6)
```
handrews, line 39: If instead you say "fragment-only URI using URI-encoded JSON Pointers per..." then you can get rid of the statement about parsing order further down, because fragment-only URIs, full URIs, JPs, and RJPs do not overlap at all in syntax.  Unless you're trying to support some non-fragment-only references?  But that seems confusing.

json-everything.net/wwwroot/md/vocabs-data-2022.md
```diff
@@ -16,66 +16,76 @@ The intent for this keyword is to cover the use cases discussed across several i
...
+Because JSON Pointers and Relative JSON Pointers are syntactically valid URIs, the value MUST be checked in the sequence indicated above in order to properly identify the pointer types from other URIs.
 
-If any single reference cannot be resolved, validation MUST fail; otherwise the full value at the specified location is returned.
+For each successfully resolved reference, the full value at the specified location MUST be returned.
 
-If the resolved value is not valid for the associated keyword, validation MUST fail.
+If a reference cannot be resolved, or if a resolved value is not valid for the associated keyword, evaluation MUST halt.  Implementations SHOULD use native features of their language to report the failure as appropriate.  Implementations MAY continue to attempt to resolve other references so that multiple resolution failures can be reported together, however further schema evaluation MUST NOT continue.
```
handrews, line 74: Interesting - I had not entirely put this together with the `$ref` stuff.

I do think I agree with it.  Were this purely about instance references, you could classify the failure as a validation failure (e.g. this keyword requires two parts of the instance to exist in some sort of relation, and since they don't that is a validation failure because it's all about the instance).  But the external data aspect means that unresolved references are errors, not validation failures.

json-everything.net/wwwroot/md/vocabs-data-2022.md
```diff
@@ -16,66 +16,76 @@ The intent for this keyword is to cover the use cases discussed across several i
...
 
-### 3.4 Errors
+#### 4.3.1 External Data Access
 
-The `data` keyword can generate two validation failure states as specified by this document:
+Implementations SHOULD provide a means to pre-load and cache any external references prior to evaluation but MAY be configured to fetch external documents at evaluation time.  Documents fetched from URIs which contain a JSON Pointer fragment MUST be interpreted using a media type that allows resolution of such fragments.
```
handrews, line 78: Probably worth appending "such as `application/schema-instance+json`" as a lot of people will not understand media types well enough to figure out what this means in practice.

json-everything.net/wwwroot/meta/data-2022
```diff
@@ -22,9 +22,9 @@
             "additionalProperties": {
                 "type": "string",
                 "oneOf": [
-                    { "format": "uri-reference" },
                     { "format": "json-pointer" },
-                    { "format": "relative-json-pointer" }
+                    { "format": "relative-json-pointer" },
+                    { "format": "uri-reference" }
```
handrews, line 27: You can add `"pattern": "^#"` to restrict this to fragment-only URIs if you go with that suggestion I made earlier.

handrews (approved):
I added a few more comments and nitpicks, but nothing that would be considered blocking even if accepted.  I'm excited to see this!

json-everything.net/wwwroot/md/vocabs-data-2022.md
```diff
@@ -16,66 +16,76 @@ The intent for this keyword is to cover the use cases discussed across several i
...
+### 4.1 Syntax and Semantics
 
 The value of `data` must be an object.  The keys of the object are interpreted as JSON Schema keywords, and the values MUST be one of
 
-- JSON Pointers per [RFC 6901](https://www.rfc-editor.org/rfc/rfc6901)
+- JSON Pointers per [RFC 6901](https://www.rfc-editor.org/rfc/rfc6901) 
 - [Relative JSON Pointers](https://json-schema.org/draft/2019-09/relative-json-pointer.html)
-- URI references per [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986), potentially with JSON Pointer fragments
+- URI-encoded JSON Pointer identifier per [RFC 6901, §6](https://www.rfc-editor.org/rfc/rfc6901#section-6)
+- Absolute URIs per [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986), optionally with a JSON Pointer fragment identifier
```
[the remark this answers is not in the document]
gregsdennis: "with" here means "additionally"

json-everything.net/wwwroot/md/vocabs-data-2022.md
```diff
@@ -16,66 +16,76 @@ The intent for this keyword is to cover the use cases discussed across several i
...
 
-`data` MUST be processed contextually in the same manner as the schema in which it is used.  Specifically,
+`data` MUST be processed contextually in the same manner as the host schema.  Specifically,
 
 - URI resolution MUST be performed in the same manner as `$ref` (per section 3.3).
-- Keys that are ignored by the parent schema MUST also be ignored by the formed subschema.
+- Keys that are ignored by the host schema MUST also be ignored by the formed schema.
 
-Implementations SHOULD validate the formed schema against the containing schema's meta-schema (as specified by `$schema`) to ensure that it is a syntactically valid schema object.
+Implementations SHOULD validate the formed schema against the host schema's meta-schema (as specified by `$schema`) to ensure that it is a syntactically valid schema object.
```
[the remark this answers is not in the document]
gregsdennis: The intent is to get the implementation to validate that the data it has found creates a valid schema.  I can update this to be more explicit in the intent.

json-everything.net/wwwroot/md/vocabs-data-2022.md
gregsdennis: Sure, I can make this update.  The only catch is I _am_ trying to support full URIs + JP fragments (`https://example.com/data#/foo`).

json-everything.net/wwwroot/md/vocabs-data-2022.md
gregsdennis: Yeah, my thinking is that the "unresolved reference" aspect trumps the "in the instance" aspect, especially because not all references are in the instance and all dereference attempts should be treated the same.

json-everything.net/wwwroot/meta/data-2022
gregsdennis: Do we have a format for full (absolute) URI that could also include a JP fragment?  I'd have to split out `uri-reference` to support "JP fragment URI" and "full URI with (+) optional JP fragment" as separate items (which is fine, but I'd have to do it).

json-everything.net/wwwroot/md/vocabs-data-2022.md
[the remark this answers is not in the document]
gregsdennis: reworded

json-everything.net/wwwroot/md/vocabs-data-2022.md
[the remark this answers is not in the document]
gregsdennis: ✔️

json-everything.net/wwwroot/md/vocabs-data-2022.md
gregsdennis: ✔️

json-everything.net/wwwroot/meta/data-2022
gregsdennis: I've split this to `iri` and `iri-reference`-with-`pattern`.

(I've also updated to use IRI instead of URI.)

handrews: > Do we have a format for full (absolute) URI that could also include a JP fragment? 

@gregsdennis one of my many half-designed `format`-replacement vocabularies would be capable of this 🙃

handrews (approved):
Latest changes look good!

PR #319, opened 2022-08-13
Fix 318: Numberify can parse floating point numbers in any culture

### LINKED ISSUE #318: Numberify doesn't work in all cultures
**Environment**
- Commit: f18e645179fb8b068a9c3183c73ffbc1782a709a (from 2022-08-12)
- Nuget Package: -
- Nuget Version: -
- OS: Windows
- .Net Target: All

**Describe the bug**
When converting a string to a number with `Numberify()`, floating point numbers are only parsed correctly if the current culture has a english-like number format. For eg. `de-AT` this isn't the case, so the input string `3.14` just returns ~the integer value `3`~ `null`,  as `,` is the decimal separator.

**To Reproduce**
```csharp
CultureInfo.CurrentCulture = new CultureInfo("de-AT");
var number = JsonValue.Create("3.14").Numberify();
```

**Expected behavior**
It should return correct floating point numbers regardless of current culture.

warappa:
Fixes #318

### DIFF
--- a/JsonLogic.Tests/GithubTests.cs
+++ b/JsonLogic.Tests/GithubTests.cs
@@ -1,4 +1,5 @@
-﻿using System.Text.Json;
+﻿using System.Globalization;
+using System.Text.Json;
 using System.Text.Json.Nodes;
 using Json.Logic.Rules;
 using Json.More;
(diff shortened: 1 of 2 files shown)


JsonLogic.Tests/GithubTests.cs
```diff
@@ -138,4 +139,13 @@ public void Issue313_VarEqualsNull()
...
 		Assert.IsTrue(result.IsEquivalentTo(true));
 	}
+
+	[Test]
+	public void Issue318_CanParseStringAsFloatingPointNumberInAnyCulture()
+	{
+		CultureInfo.CurrentCulture = new CultureInfo("de-AT");
+		var number = JsonValue.Create("3.14").Numberify();
+
+		Assert.AreEqual(3.14d, number);
```
gregsdennis, line 149: Could you wrap this in a try/finally and reset the culture to the value it was before the test ran (capture its current value before changing it and restore it in the finally block)?  That way it doesn't affect tests that run afterward.

warappa: Fixed as proposed.

PR #320, opened 2022-08-14
Logic v3.1.2

### LINKED ISSUE #318: Numberify doesn't work in all cultures

gregsdennis:
Depends on #319
Resolves #318

### DIFF
--- a/JsonLogic.Tests/GithubTests.cs
+++ b/JsonLogic.Tests/GithubTests.cs
@@ -1,4 +1,5 @@
-﻿using System.Text.Json;
+﻿using System.Globalization;
+using System.Text.Json;
 using System.Text.Json.Nodes;
 using Json.Logic.Rules;
 using Json.More;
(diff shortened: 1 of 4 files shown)


gregsdennis:
@warappa once this is merged, a new version will be published.  Thanks for your help!

PR #322, opened 2022-08-19
Fix JsonPatch move operation on arrays

z4kn4fein:
Hi there!

I've captured an issue with the `move` JsonPatch operation that causes the following:
- Given the following source data:
   ```
   {
    "items": [
        {
            "value": 1
        },
        {
            "value": 2
        },
        {
            "value": 3
        },
        {
            "value": 4
        }
     ]
   }
   ```
- Patched with the following op:
   ```
   [{
       "op": "move",
       "path": "/items/0",
       "from": "/items/1"
   }]
   ```
- Gives the following result:
   ```
   {
     "items": [
       {
         "value": 2
       },
       {
         "value": 3
       },
       {
         "value": 4
       }
     ]
   }
   ```

I've just noticed that around v2.0.0, an array `Insert()` operation was changed to an indexed write, causing this "deleting shift" effect. I switched back to `Insert()` and added unit tests to cover my changes. Let me know if I should do anything differently. Thanks!

### DIFF
--- a/JsonLogic.Tests/GithubTests.cs
+++ b/JsonLogic.Tests/GithubTests.cs
@@ -1,4 +1,5 @@
-﻿using System.Text.Json;
+﻿using System.Globalization;
+using System.Text.Json;
 using System.Text.Json.Nodes;
 using Json.Logic.Rules;
 using Json.More;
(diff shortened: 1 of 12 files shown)


gregsdennis:
Thanks for this. I'll investigate the change you mentioned in 2.0.0 to find out why that occurred.  I think this change is good.

gregsdennis:
Yeah, I'm not sure where the bug was introduced, but you're right about the behavior.  I'll set up a release branch and have this pushed out soon.

Thanks again.

PR #359, opened 2022-12-16
Schema/better access

### LINKED ISSUE #357: Better access into schema contents
Users could benefit from being able to access the data contained in the schema a bit more easily.

Add an extension `.Get<T>() where T : IJsonSchemaKeyword` to get the indicated keyword.

This can open the door to a set of extensions that directly get the values of the spec-defined keywords, e.g. `.GetTitle()`.

---

Additionally, I would like to explore using a dictionary to store the keywords by name instead of in a list.  (The `Keywords` property would be a forwarded property to the `Values` collection on the keyword, so no breakages.)

This would make accessing keywords faster.

gregsdennis:
Resolves #357
Relates to https://github.com/microsoft/OpenAPI.NET/issues/795

@darrelmiller please see if this moves toward resolving schema data (per our conversation in Slack) in a way that will help you.  Most notably, [these extensions](https://github.com/gregsdennis/json-everything/pull/359/files#diff-8d8116f7f57bd221fa3fa3f7b850b2f80fe4d5bbc99a794ba84608893e87dd68) (and imagine more like them or build yourself as necessary).

### DIFF
--- a/JsonSchema.Data/JsonSchemaExtensions.cs
+++ b/JsonSchema.Data/JsonSchemaExtensions.cs
@@ -2,8 +2,14 @@
 
 namespace Json.Schema.Data;
 
+/// <summary>
+/// Some extensions for <see cref="JsonSchema"/>
+/// </summary>
 public static class JsonSchemaExtensions
 {
+	/// <summary>
+	/// Gets the references in `data` if the keyword exists.
+	/// </summary>
 	public static IReadOnlyDictionary<string, IDataResourceIdentifier>? GetData(this JsonSchema schema)
 	{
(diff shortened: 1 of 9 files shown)


darrelmiller:
Thank you for this.  We are going to build some examples of code that leverage these extensions and get feedback from our team.

gregsdennis:
This would have been such a great use case for the [proposed extension properties C# feature](https://github.com/dotnet/csharplang/discussions/5811).  (I'm pretty sure that it almost made it into C# 8, but I couldn't find that.)

gregsdennis:
I think there's one test that will fail.  It's just the test suite not being up-to-date.  I've fixed it in #360.

PR #360, opened 2022-12-17
Possible approach to support schemas nested in non-schema docs

### LINKED ISSUE #355: Reference non-schema document root
## Problem

This implementation doesn't handle non-schema roots.  This means that references that expect to be resolved from the open api root, can't be resolved.

## Hack

I had a (probably hacky/bad) idea of how to resolve this: implement the open api doc type as a schema and its properties (at least the ones that hold schemas that need to be referenced) as keywords.

I'm not expecting this to go anywhere; just wondered if it'd work, and it kinda does.  I think I'd rather implement some mechanism where you can imbed a schema in a non-schema.

### (Re-)Introduce `IRefResolvable` or similar

This solution would be to provide an interface that can be implemented that defines what's required to resolve references.  Then JsonSchema implements that and the open api doc type can implement it as well.  (I actually just got rid of IRefResolvable with my update to v4 because I figured out how to do all that resolving internally.) 

### Update `SchemaRegistry` to handle non-schema roots

SchemaRegistry currently expects documents to be a JSON Schema.  This doesn't necessarily _need_ to be the case.  Just need to figure out how to do this and not break everything else.

gregsdennis:
Resolves #355 
Relates to https://github.com/microsoft/OpenAPI.NET/issues/795

There are some comments in the code, but basically the `SchemaRegistry` now tracks `IBaseDocument`s instead of just `JsonSchema`s.  The `OpenApiDoc` (or whatever it'll be called) will need to:

- implement `IBaseDocument`
- register itself under a base URI
- perform an initialization scan to discover the schemas it contains
- set the document's base URI on every contained schema

It's a bit hacked together right now, but it works.

### DIFF
--- a/JsonSchema/RecursiveRefKeyword.cs
+++ b/JsonSchema/RecursiveRefKeyword.cs
@@ -55,11 +55,15 @@ public class RecursiveRefKeyword : IJsonSchemaKeyword, IEquatable<RecursiveRefKe
 			if (scopeRoot == null)
 				throw new Exception("This shouldn't happen");
 
-			if (scopeRoot.RecursiveAnchor == null) continue;
+			if (scopeRoot is not JsonSchema schemaRoot)
+				throw new Exception("Does OpenAPI use anchors?");
 
-			if (targetBase.RecursiveAnchor == null) break;
+			if (schemaRoot.RecursiveAnchor == null) continue;
 
-			targetSchema = scopeRoot.RecursiveAnchor;
+			if (targetBase is JsonSchema targetBaseSchema &&
+			    targetBaseSchema.RecursiveAnchor == null) break;
+
+			targetSchema = schemaRoot.RecursiveAnchor;
 			break;
 		}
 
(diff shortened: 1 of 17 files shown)


JsonSchema/RecursiveRefKeyword.cs
```diff
@@ -55,11 +55,15 @@ public void Evaluate(EvaluationContext context)
 			if (scopeRoot == null)
 				throw new Exception("This shouldn't happen");
 
-			if (scopeRoot.RecursiveAnchor == null) continue;
+			if (scopeRoot is not JsonSchema schemaRoot)
+				throw new Exception("Does OpenAPI use anchors?");
```
gregsdennis, line 59: Not sure this is entire relevant here since `$recursiveRef` is exclusively 2019-09.

darrelmiller:
Maggie and I discussed this today.  This looks like an approach that would work for us.  We will do some more investigation on our side.

gregsdennis:
I think this is going to be a good addition to the library, so I'm going to go ahead and incorporate it into the beta.

Issue #368, opened 2023-01-19
Could json logic operators be protected interal? 


dcook-net:
Hi, 

I was wondering if there was a technical reason for marking the Json Logic operators as internal?

We are trying to extend the library a little for internal use, and would like to override the behaviour of several comparison operators in certain circumstances, so that if the data being operated on meets a certain condition, ie, it looks like a Date, then we perform our own comparison, otherwise, we defer to the code that already exists in say <, >, =>, =< for example.

Currently, we can't inherit from or instansiate these operators, without maybe resorting to reflection.

Would you be open to the idea of changing the access modifiers on operators, and thier fields to protected internal?

I can raise a PR for this work if it's something that works for you.
If this is not desirable for you, can you suggest an alternative?

For reference, here is a sample of what we are trying to achieve;

```c#

[Operator(">")]
public class CustomMoreThanRule : Json.Logic.Rules.MoreThanRule
{
     public CustomMoreThanRule(Rule a, Rule b) : base(a, b)
     {
     }

     public override JsonNode? Apply(JsonNode? data, JsonNode? contextData = null)
     {
         var a = A.Apply(data, contextData);
         var b = B.Apply(data, contextData);

         // ignoring for now what happens when only one is a date!
         if (a.IsDateValue() && b.IsDateValue())
             return CompareDates(a, b);

         return base.Apply(data, contextData);
    }
}

```

Thanks,
Dave

gregsdennis:
> I was wondering if there was a technical reason for marking the Json Logic operators as internal?

It's more a philisophical reason.  The primary mechanism for creating custom behavior is through [defining custom rules](https://json-everything.net/json-logic#creating-new-operators).  If you register an operation that already exists, it'll overwrite with your rule.

However, I can see an argument for wanting to actually override the rule class in order to take advantage of the default behavior except for certain cases.

I was going to suggest a composition approach for now, but that requires that you be able to access the `MoreThanRule` constructor, which is also internal.

Yeah, I think `internal protected` is a good option.  Give me a couple days, and I'll have an update out.

dcook-net:
I agree that composition approach would be preferable, though that would require making the classes and their constructors public.
You could also enforce composition over inheritance by sealing the operator classes.

dcook-net:
Created a PR for the internal protected approach.
Happy to apply changes if this isn't what you were thinking.

Cheers,
Dave

gregsdennis:
[...]

PR #369, opened 2023-01-21
Make operators in the specification private protected to allow inheritance.

dcook-net:
Resolves https://github.com/gregsdennis/json-everything/issues/368

### DIFF
--- a/JsonLogic/Rules/BooleanCastRule.cs
+++ b/JsonLogic/Rules/BooleanCastRule.cs
@@ -12,9 +12,16 @@ namespace Json.Logic.Rules;
 [JsonConverter(typeof(BooleanCastRuleJsonConverter))]
 public class BooleanCastRule : Rule
 {
-	internal Rule Value { get; }
+	/// <summary>
+	/// The value to test.
+	/// </summary>
+	protected internal Rule Value { get; }
 
-	internal BooleanCastRule(Rule value)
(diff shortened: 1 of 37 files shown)


JsonLogic/Rules/BooleanCastRule.cs
```diff
@@ -30,8 +37,6 @@ internal BooleanCastRule(Rule value)
 	/// <returns>The result of the rule.</returns>
 	public override JsonNode? Apply(JsonNode? data, JsonNode? contextData = null)
 	{
-		var value = Value.Apply(data, contextData);
```
dcook-net, line 33: This wasn't being used.

gregsdennis: Oh, I probably had that in there as a debug so I could see what the result was.

gregsdennis:
I really need to fix that labeler build.  It's failing because of permissions across forks.

gregsdennis:
@dcook-net can you try rebasing on my `logic/rule-accessibility` branch please?  I don't expect there to be any conflicts.

dcook-net:
@gregsdennis 👍🏻 will do this today.

gregsdennis:
Thanks for this.  I'll merge it into the branch and tweak a few things.  When I merge that into `master` it'll publish.

Issue #373, opened 2023-02-01
Operators '<' and '<=' throws errors in case 3 parameters were passed


alexkharuk:
Hello! 

**Environment**
- Nuget Package: JsonLogic
- Nuget Version: 3.3.0
- OS: macOS
- .Net Target: net6.0

**Describe the bug**
Operators '<' and '<=' throws errors in case 3 parameters were passed and one of them or all have type String, like "1". But parameters are working fine in case 2 parameters were passed

**To Reproduce**
{"<" : [1, 2]} - passed
{"<" : ["1", 2]} - passed
{"<" : [1, 2, 3]} - passed
{"<" : ["1", 2, 3]} - throw an error "Lower bound must be a number"

i checked your code and found that:

in case 2 parameters you are trying to convert strings to numbers:
`var numberA = a.Numberify();`

but in case 3 parameters you are not:
`var low = (A.Apply(data, contextData) as JsonValue)?.GetNumber();`

gregsdennis:
Thanks for finding this.  I'm happy to accept a PR to fix.

alexkharuk:
Hello. PR was created: https://github.com/gregsdennis/json-everything/pull/375

PR #375, opened 2023-02-02
Update between operator

alexkharuk:
issue link -> https://github.com/gregsdennis/json-everything/issues/373

Between operator can handle operands that can be parse to a number.
Numberify values in case of 3 parameters passed into LessThenRule, LessThenEqualRule


gregsdennis:
Thank you.  Have you been able to test these cases in jsonlogic.com?

Their test suite is really basic and should cover cases like this, but it doesn't.

gregsdennis:
The changes look good.  Thanks.  I'll create a new branch so I can do versioning things and update this PR to merge into that.

PR #378, opened 2023-02-08
docs: fix Build badge in README

amis92:
I've just noticed the badge doesn't render correctly, so I went and copied the current markdown from GH Actions tab.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 <br>
 
-[![Build status](https://github.com/gregsdennis/json-everything/workflows/.NET%20Core/badge.svg)](https://github.com/gregsdennis/json-everything/actions?query=workflow%3A%22Build+%26+Test%22)
+[![Build & Test](https://github.com/gregsdennis/json-everything/actions/workflows/dotnet-core.yml/badge.svg?event=push)](https://github.com/gregsdennis/json-everything/actions/workflows/dotnet-core.yml)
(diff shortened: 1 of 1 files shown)


gregsdennis:
Thanks.  I have a separate badge for test results.  Can you look into removing that from this one so that it just gives build status?

amis92:
@gregsdennis it's just a fix because the old URL was broken. GitHub workflow badges don't support filtering on jobs inside, it'd need to be a custom badge as it is with tests.

gregsdennis:
Ah, okay...  It looks like the title comes from the `name` specified at the top of the GH action .yml file.  I don't need the event spec, though.  That can be removed.

amis92:
Depends what you want to see on the badge. Without the filter, it'll show latest run, for example for a pull request.

README.md
```diff
@@ -4,7 +4,7 @@
 
 <br>
 
-[![Build status](https://github.com/gregsdennis/json-everything/workflows/.NET%20Core/badge.svg)](https://github.com/gregsdennis/json-everything/actions?query=workflow%3A%22Build+%26+Test%22)
+[![Build & Test](https://github.com/gregsdennis/json-everything/actions/workflows/dotnet-core.yml/badge.svg?event=push)](https://github.com/gregsdennis/json-everything/actions/workflows/dotnet-core.yml)
```
amis92, line 7:
```suggestion
[![Build & Test](https://github.com/gregsdennis/json-everything/actions/workflows/dotnet-core.yml/badge.svg?branch=master&event=push)](https://github.com/gregsdennis/json-everything/actions/workflows/dotnet-core.yml)
```

gregsdennis:
Thanks for fixing this.

PR #399, opened 2023-03-09
use trx test logger

gregsdennis:
Dropping JUnit logger because of https://github.com/spekt/junit.testlogger/issues/59


gregsdennis:
The analysis of the test report files takes longer than actually running and writing the tests, but all of this works.

PR #403, opened 2023-03-14
added basic bundling support

### LINKED ISSUE #401: Creating Compound Schema Document (bundle/bundling)
**Environment**
- Nuget Version: 3.3.2

**How can I help you today?**

I want to create Compound Schema Documents from a set of schemas stored in separate files. 
I think this is currently not possible using the JsonSchema.Net package, and I'd like it to be :D If it is, please advise how.

The bundling is described on [json-schema.org](https://json-schema.org/understanding-json-schema/structuring.html#bundling).

**Additional context**

I've browsed the issues and I suspect that having #343 implemented would allow me to implement a crude bundler myself, and also generate the C# code myself a bit easier as I'm doing a lot of customizations.

However, a direct Bundler/Compound Schema Document builder provided by this project would be much more helpful.

gregsdennis:
Resolves #401

### DIFF
--- a/JsonSchema.Tests/BundlingTests.cs
+++ b/JsonSchema.Tests/BundlingTests.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using System.Text.Encodings.Web;
 using System.Text.Json;
 using NUnit.Framework;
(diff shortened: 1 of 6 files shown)


gregsdennis:
@amis92 I'm going to proceed with this.  I think it's good.  If you want to extract the individual schemas and bundle them a different way, you can just get them all from the `$defs` keyword.

amis92:
@gregsdennis thanks! Also about getting just the list of schemas - good idea with just accessing the `$defs` :)

PR #413, opened 2023-03-24
date-time validation check using regex

faldor20:
DateTimes that have very high precision eg: 2023-03-20T05:41:23.802351665Z can be parsed by dotnet and are allowed by jsonschema but don't get accepted by TryParseExact
This uses regex as a fallback to check a datetime just incase it's actually compliant event though TryParseExact disagrees.

### DIFF
--- a/JsonSchema.Tests/FormatTests.cs
+++ b/JsonSchema.Tests/FormatTests.cs
@@ -33,6 +33,18 @@ public class FormatTests
 		Assert.False(result.IsValid);
 	}
 
+	[Test]
+	public void longDateTime_pass()
+	{
+		JsonSchema schema = new JsonSchemaBuilder()
+			.Format(Formats.DateTime);
+
+		var value = JsonNode.Parse("\"2023-03-22T07:56:28.610645938Z\"");
+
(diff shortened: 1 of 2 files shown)


gregsdennis (changes requested):
It looks like (line 19) that I only supply formats up to 7-digit precision.  Have you tried adding the additional digits here instead of regex?

Also can you add a unit test to cover this case, please?

faldor20:
I did try that, unfortunately adding more does nothing. 
https://learn.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings
It seems the list here is comprehensive. 
I couldn't see where you were testing it in the first place so I wasn't quite sure where to add a test.
If you point me in the right direction I'll happily do so though :)

Thanks for getting back to me so quick!

gregsdennis:
> I couldn't see where you were testing it in the first place so I wasn't quite sure where to add a test.

Please add to `FormatTests`.

gregsdennis:
Thanks.  I'm going to merge this into #418 just to combine releases.

gregsdennis (approved):

PR #414, opened 2023-03-24
Support local refs

faldor20:
I have a jsonschema project that uses local refs. However because of the way that context.Scope.localScope changes when entering allOf anyOf oneOf etc type blocks. $ref won't resolve them based on baseUri. 
This adds a fallback method of resolution that just attempts to resolve against the baseUri ignoring any of the internal references.

Feel free to complain If this is really stupid. It did solve my problem though.


gregsdennis (changes requested):
It's not clear what the problem you're trying to solve is.  Can you open an issue where we can discuss, please?

PR #418, opened 2023-03-26
lots of minor updates

### LINKED ISSUE #406: Get (valid!) JsonPath or JsonPointer from a JsonNode
**Environment**
- Nuget Package: JsonPath.Net
- Nuget Version: 0.4.0
- OS: dotnetFiddle/Linux
- .Net Target: 

I'd like to be able to generate a correct JSON Path and/or JSON Pointer from a System.Text.Json.Nodes.JsonNode type.

**Describe alternatives you've considered**

Post a bug report to dotnet/runtime.

**Additional context**

Json path generated by BCL's `JsonNode.GetPath()` for a property like `$id` is invalid and fails to be parsed by JsonPath.

I'd like to have some (extension?) methods that provide a valid JsonPath and JsonPointer given a JsonNode. It seems to already be mostly done within the code of this library, but is not available via public API. It can return a plain string, or a strongly typed parsed instance of path or pointer.

A dotnet fiddle to illustrate the issue: https://dotnetfiddle.net/vkcQQ8

### LINKED ISSUE #417: `schema.Bundle()` fails with NullReferenceException for ItemsKeyword
**Environment**
- Nuget Version: 4.0.0
- OS: Linux
- .Net Target: .NET 7.0

**Describe the bug**

`JsonSchemaExtensions.Bundle(EvaluationOptions)` throws `NullReferenceException` given a schema containing `ItemsKeyword`.

**To Reproduce**

```csharp
#r "nuget: JsonSchema.Net, 4.0.0"
using Json.Schema;
// items: true
var singleItemSchema = new JsonSchemaBuilder()
    .Items(JsonSchema.True)
    .Build();
singleItemSchema.Bundle(); // throws
// items: [true, true]
var multiItemSchema = new JsonSchemaBuilder()
    .Items(new[]{JsonSchema.True, JsonSchema.True})
    .Build();
multiItemSchema.Bundle(); // throws
```

<details>
<summary>stacktrace</summary>

for `singleItemSchema`:

```
System.NullReferenceException: Object reference not set to an instance of an object.
at System.Linq.Enumerable.SelectManySingleSelectorIterator`2.MoveNext()
at System.Collections.Generic.List`1.InsertRange(Int32 index, IEnumerable`1 collection)
at Json.Schema.JsonSchemaExtensions.Bundle(JsonSchema jsonSchema, EvaluationOptions options)
```

for `multiItemSchema`:

```
System.NullReferenceException: Object reference not set to an instance of an object.
at Json.Schema.JsonSchemaExtensions.Bundle(JsonSchema jsonSchema, EvaluationOptions options)
```

</details>

**Expected behavior**

Method returns bundled schema without exception.

**Additional context**

It seems (at least for the single-item case) this code is the problem:

https://github.com/gregsdennis/json-everything/blob/44d5fe8f83882d061b336cedc792bcc60563fd3e/JsonSchema/JsonSchemaExtensions.cs#L122-L125

On line 123, `ItemsKeyword` is `ISchemaCollector`, but when initialized with a single non-array value, it will return a null for `ISchemaCollector.Schemas`.

For multi-item case, I am not sure.

gregsdennis:
Resolves #417 
Resolves #406 (relates to https://github.com/dotnet/runtime/issues/83547)


JsonSchema/JsonSchemaExtensions.cs
```diff
@@ -119,10 +119,15 @@ public static JsonSchema Bundle(this JsonSchema jsonSchema, EvaluationOptions? o
...
-				schemasToSearch.AddRange(schema.Keywords.OfType<ICustomSchemaCollector>().SelectMany(x => x.Schemas));
+				// I shouldn't ordinarily have to check for null here,
+				// but ItemsKeyword implements two interfaces and only one
+				// will be populated at a time.  There may be others
+				// that implement multiple in the same way, so I need to
+				// consider that.
+				schemasToSearch.AddRange(schema.Keywords.OfType<ISchemaContainer>().Where(x => x.Schema != null!).Select(x => x.Schema));
+				schemasToSearch.AddRange(schema.Keywords.OfType<ISchemaCollector>().Where(x => x.Schemas != null!).SelectMany(x => x.Schemas));
+				schemasToSearch.AddRange(schema.Keywords.OfType<IKeyedSchemaCollector>().Where(x => x.Schemas != null!).SelectMany(x => x.Schemas.Values));
+				schemasToSearch.AddRange(schema.Keywords.OfType<ICustomSchemaCollector>().Where(x => x.Schemas != null!).SelectMany(x => x.Schemas));
```
amis92, line 130: A similar issue can occur in:

https://github.com/gregsdennis/json-everything/blob/44d5fe8f83882d061b336cedc792bcc60563fd3e/JsonSchema/JsonSchema.cs#L324-L350

That's why I suggested having a single place for such an iteration.

gregsdennis: It can't happen here at least not with `items`.

This switch only runs one branch, and it evaluates them in this order.  If the `items` keyword is a schema, it'll hit the first branch and bypass the second.  If it's an array, it'll bypass the first branch and hit the second.

The code in `.Bundle()` however runs all of the cases.  This was the problem.

amis92: Okay, so this code is perfectly fine - why not use it as well in `.Bundle()`? Instead of 4x `schemasToSearch.AddRange(schema.Keywords.OfType...)`, there'd be just `schemasToSearch.AddRange(schema.Keywords.SelectMany(JsonSchema.GetSubschemas))` - it just needs to be made `internal` instead of `private`.

gregsdennis: Interesting.  I'll have a look.

gregsdennis: 10 points to @amis92 for preventing me from re-inventing the wheel.

Issue #426, opened 2023-04-04
Internal references don't work for schemas with a "file://" type baseUri


faldor20:
**Environment**
- Nuget Package: JsonSchema.Net
- Nuget Version: . 4.02
- OS: windows
- .Net Target: net6.0

**Describe the bug**
When evaluating a schema with a base uri that is path internal references don't work

**To Reproduce**
I just modified your existing openapi test by moving the schema into a file 
```csharp
var openApiDocText = File.ReadAllText("./schema.json");
var openApiDocNode = JsonNode.Parse(openApiDocText);
var openApiDoc = new OpenApiDoc(new Uri(System.IO.Path.GetFullPath("./")), openApiDocNode!);
var targetSchema = openApiDoc.FindSubschema(JsonPointer.Parse("/webhooks/newPet/post/requestBody/content/application~1json/schema"), EvaluationOptions.Default);

var payload = new JsonObject
{
	["id"] = 1,
	["name"] = "fido",
	["tag"] = "a very good dog"
};

var result = targetSchema!.Evaluate(payload);
```

**Expected behavior**
This should work even with file baseUris

**Additional context**

I fixed this by adding this to refKeyword.cs
```csharp
var newUri = new Uri(context.Scope.LocalScope, Reference);
string fragment = newUri.Fragment;
//If the uri is a file we set the fragment here because it will be lost in the uri
if (context.Scope.LocalScope.IsFile && Reference.OriginalString.StartsWith("#"))
{
	newUri = context.Scope.LocalScope;
	fragment = Reference.OriginalString;
}
...
...
if (JsonPointer.TryParse(fragment, out var pointerFragment))
..
..

```
The issue is caused by file URIs not having support for fragments so the # gets converted to %23 and the uri cannot be resolved. 
This fixes that by keeping the fragment seperate in file URIs

gregsdennis:
Thanks for reporting this.  I don't think I ever considered that file URIs would disallow fragments.

Would you be willing to post a PR for this?

- Separate commits that:
    - replicate the problem with a unit test that fails
    - fix the problem (make the test pass)
- Merge to branch `schema/file-refs-with-pointers`

nvkgbr:
Was this change reverted? I am facing simillar issues currently. 

I have a reference like this:
 `    "yProperty": {
      "$ref": "#/definitions/property"
    }
`

And I am gettin an exception: 

`
[...]

PR #427, opened 2023-04-04
File fragment refs

faldor20:
Fix file fragment refs and add a test 
Resolves https://github.com/gregsdennis/json-everything/issues/426


faldor20:
Locally the tests all work except for the one on line 116 of validation.cs. Which also fails on master

gregsdennis:
Very rarely, something goes flaky.  I'm just going to re-run the job.  It's fine on my local as well.

PR #430, opened 2023-04-13
Further file fragment fixes

faldor20:
Okay,  so I recently realised my file fragment fix still breaks sometimes. If you reference a file like so:
```json
{
"$ref":"./potato.json#/definitions/leek"
}
```
it doesn't detect that it's a fragment, then joins it to the base fileURI and breaks 

This attempts to fix that. Currently It uses the most simple approach possible:
Split at a hash and error if there is more that one because we cannot tell what is fragment and what isn't

### Ideas to detect fragments better and support paths with hashes in them:
- anything after the last # if the uri doesn't end in a .json 
- any thing after a block of text that looks like '.json#/'
- 
Both of these are somewhat flawed. They don't work if you have combinations of the following:

- File paths and names can have hashes.
- files with json in them don't necessarily need to end in .json 

What are your thoughts on this? Maybe you can think of a better way.

### DIFF
--- a/JsonSchema.Tests/ReferenceTests.cs
+++ b/JsonSchema.Tests/ReferenceTests.cs
@@ -1,0 +1,37 @@
+using System.IO;
+using System.Text.Json.Nodes;
+using NUnit.Framework;
+
+namespace Json.Schema.Tests;
+
+public class References
+{
+	
+	private static string GetFile( string name)
+	{
+		return Path.Combine(TestContext.CurrentContext.WorkDirectory, "Files","Referencing", $"{name}.json")
+			.AdjustForPlatform();
+	}
+
+	private static string GetResource( string name)
+	{
+		return File.ReadAllText(GetFile( name));
+	}
+
+	[Test]
+	public void ReferenceFragmentFromFile()
+	{
+		var baseSchema =JsonSchema.FromFile(GetFile("base_schema"));
+		var refSchema =JsonSchema.FromFile(GetFile("ref_schema"));
+		
+		var baseData = JsonNode.Parse( GetResource("base_data"));
+		
+		SchemaRegistry.Global.Register(refSchema);
+		SchemaRegistry.Global.Register(baseSchema);
+
+		var res=baseSchema.Evaluate(baseData);
+		
+		res.AssertValid();
+	}
+	
+}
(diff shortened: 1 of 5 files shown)


JsonSchema/RefKeyword.cs
```diff
@@ -49,12 +49,25 @@ public void Evaluate(EvaluationContext context)
...
+				fragment = Reference.OriginalString;
+				newUri = context.Scope.LocalScope;
+			}
+
+			else
+			{
+				var parts=Reference.OriginalString.Split('#');
+				if (parts.Length != 2)
+					throw new JsonSchemaException(
+						$"Given a reference with more than one '#' in it. We cannot tell if this is a fragment, or where the fragment starts. Please don't use '#'s in fileNames or paths.\n Reference:{Reference.OriginalString} ");
```
gregsdennis, line 66: This is rather long.  I'd prefer a shorter error message with a longer explanation in the docs.

```suggestion
						"File references with multiple hashes are not supported.");
```

gregsdennis (changes requested):

JsonSchema.Tests/ReferenceTests.cs
```diff
@@ -0,0 +1,37 @@
+using System.IO;
+using System.Text.Json.Nodes;
+using NUnit.Framework;
+
+namespace Json.Schema.Tests;
+
+public class References
+{
+
```
gregsdennis, line 9: Please remove this blank line.

JsonSchema.Tests/ReferenceTests.cs
```diff
@@ -0,0 +1,37 @@
...
+		var baseData = JsonNode.Parse( GetResource("base_data"));
+		
+		SchemaRegistry.Global.Register(refSchema);
+		SchemaRegistry.Global.Register(baseSchema);
+
+		var res=baseSchema.Evaluate(baseData);
+		
+		res.AssertValid();
+	}
+
```
gregsdennis, line 36: And this one.

JsonSchema.Tests/ReferenceTests.cs
gregsdennis, line 1: Can you add a test to check that the exception is thrown, please?

gregsdennis (changes requested):

faldor20:
Oh, haha, well I was just about to push my changes but If you've got a fix that's great. Hilarious timing though.
Do you want to integrate the unit tests I included into your branch, just to ensure your change works for my case?

gregsdennis:
Yes, that'd be great. Thanks.

PR #437, opened 2023-04-23
Tests for the # file uri problem

faldor20:
This is just the tests from my previous PR #430


gregsdennis:
Thanks for adding these. I'm interested what the error is for the one that throws.  I'm guessing it can't resolve the reference, but I'm curious what it's trying to resolve.

faldor20:
It's trying to resolve a reference to a fragment within a file who's path contains a # so you get the "two hashes in a path" fragment ambiguous issue.

gregsdennis:
Yeah, I just need to see the error locally.  You had code for that explicitly, but that's not there now.

It's neat that the error is still thrown, though. I just need to run it.

gregsdennis:
So it looks like the fragment becomes `#_in_uri.json#/definitions/myDate`.  That's what the `Uri` class is returning, too.

I think it's probably a good idea to add something about file names with `#` into the docs, but I don't think we need to do more than that.

Issue #439, opened 2023-04-24
Date-time format should use ISO 8601-1:2019 (same as System.Text.Json)


cdecroix:
**Environment**
- Nuget Package: JsonSchema.Net
- Nuget Version: 4.0.3
- OS: Windows
- .Net Target: .Net 6.0

**Describe the bug**
Values like  "2023-03-20 05:41:23.3881075" ("T" missing between the date and time) are valid in JsonSchema but it can't be deserialized by System.Text.Json.

According to the Microsoft documentation, they parse date using the ISO 8601-1:2019 format. 
The regex added to JsonSchema in v4.0.1 is based on the previous version of that ISO.

**Expected behavior**
Validation should return an error.

gregsdennis:
The JSON Schema spec, [section 7.3.1](https://json-schema.org/draft/2020-12/json-schema-validation.html#name-dates-times-and-duration) states RFC 3339.  That's what this library uses.

cdecroix:
Is it possible to override the format validation to match with System.Text.Json?

gregsdennis:
You can create your own `format` implementation and override the default implementation.

The instructions are [here](https://json-everything.net/json-schema#schema-format).

Also, you may want to read that whole section, as it describes how to enable format validation.

Lastly, please be aware that providing date/time in a different format probably won't be compatible with other implementations, and you'll need to do this modification wherever you use this schema.

gregsdennis:
Relevant: https://ijmacd.github.io/rfc3339-iso8601/

gregsdennis:
This issue led me to look at the regex that was contributed to evaluate high-precision formats, and I discovered that that actually uses an ISO 8601 regex, which is wrong.  I'll be updating that.

I'm also opening an issue with .Net for deserialization support.  I wouldn't be surprised if they come back with, "You can create a custom converter for `DateTime`."  I'm not sure if that solves your issue, though.

gregsdennis:
> Is it possible to override the format validation to match with System.Text.Json? - @cdecroix 

Yes, my library supports overriding formats, but data will not be valid according to JSON Schema (as a spec).  My library would validate it, but the same data wouldn't validate in other implementations or languages (e.g. Python).

cdecroix:
Thanks for the clarification on which spec should be used! 

For now, I overrode the validation in this lib to keep the same behavior as before (pre v4.0.1).

I did it by creating a ISchemaRefiner class that replace the format intent to use my custom validation method.
[...]

PR #442, opened 2023-04-28
fix date-time format regex to support rfc3339

gregsdennis:
Relates to #439


gregsdennis:
Happily, this seems to have improved my optional test coverage!

PR #466, opened 2023-06-02
Generating conditional schemas

gregsdennis:
This adds the ability to create conditional schemas.  Documentation will follow in a docs repo PR.


gregsdennis:
Just trying to determine if these changes should be considered breaking...  I don't think so.  I didn't have to update any of the other tests, so they didn't break, but I need to look at everything.

gregsdennis:
I think it's fine to keep this as a minor update.  There is `AdditionalItemsIntent.GetContexts()` that's being removed, but it's a leftover from a previous design, none of the other attributes have that method anymore, and I really don't expect anyone to be using it.

PR #467, opened 2023-06-04
exposing primary portions of model; makes exposed API more consistent

### LINKED ISSUE #465: Why JsonPath.Segments is internal?
**Environment**
- Nuget Package: JsonPath.Net
- Nuget Version: 0.6.1
- OS: macOS
- .Net Target: .NET 8.0.100-preview.3.23178.7

**How can I help you today?**
I need to get last path segment name, like:
- `some.test` => `test`
- `some[0].second.test2` => `test2`

As I found I could get it from `Node.Location.Segments[^1].Selectors[0].Name`, but Node.Location.Segments is `internal`.

P.S. NameSelector seems to be internal too, can you make it public?

**Additional context**

gregsdennis:
Resolves #465

@qt-kaneko, I think this is the level of exposure I'm comfortable with.  Let me know if this suits your needs.

### DIFF
--- a/JsonPath/Expressions/BooleanResultExpressionNode.cs
+++ b/JsonPath/Expressions/BooleanResultExpressionNode.cs
@@ -1,11 +1,10 @@
 ﻿using System.Diagnostics.CodeAnalysis;
 using System;
-using System.Text;
 using System.Text.Json.Nodes;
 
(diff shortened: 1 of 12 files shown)


qt-kaneko:
Yep, this includes all that I wanted, thank you.

PR #470, opened 2023-06-09
sweden and norway resources

### LINKED ISSUE #457: Localization for Sweden
**How can I help you today?**
Hi, we need validation errors localized for sweden, the same way it is done for es culture. If I were to provide translation would you be able to add support for it?

 I saw similar issue over here:
https://github.com/gregsdennis/json-everything/issues/298

pawel-wachalski:
Resolves #457 Adding resources for Sweden and Norway

### DIFF
--- a/JsonSchema/Localization/Resources.nb-no.resx
+++ b/JsonSchema/Localization/Resources.nb-no.resx
@@ -1,0 +1,213 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<root>
+  <!-- 
+    Microsoft ResX Schema 
+    
+    Version 2.0
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
+    associated with the data types.
+    
+    Example:
+    
+    ... ado.net/XML headers & schema ...
+    <resheader name="resmimetype">text/microsoft-resx</resheader>
+    <resheader name="version">2.0</resheader>
+    <resheader name="reader">System.Resources.ResXResourceReader, System.Windows.Forms, ...</resheader>
+    <resheader name="writer">System.Resources.ResXResourceWriter, System.Windows.Forms, ...</resheader>
+    <data name="Name1"><value>this is my long string</value><comment>this is a comment</comment></data>
+    <data name="Color1" type="System.Drawing.Color, System.Drawing">Blue</data>
+    <data name="Bitmap1" mimetype="application/x-microsoft.net.object.binary.base64">
+        <value>[base64 mime encoded serialized .NET Framework object]</value>
+    </data>
+    <data name="Icon1" type="System.Drawing.Icon, System.Drawing" mimetype="application/x-microsoft.net.object.bytearray.base64">
+        <value>[base64 mime encoded string representing a byte array form of the .NET Framework object]</value>
+        <comment>This is a comment</comment>
+    </data>
+                
+    There are any number of "resheader" rows that contain simple 
+    name/value pairs.
+    
+    Each data row contains a name, and value. The row also contains a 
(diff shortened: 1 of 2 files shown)


gregsdennis:
Thanks for these.  I'll get them in and published.

(I want to publish them as separate packages, but it seems that the Spanish resources are being added to the main package anyway.  I'll try to figure that out as well.)

gregsdennis:
@pawel-wachalski I've merged this for now.  I'll need to set up a build to publish it.

gregsdennis:
JsonSchema.Net v4.1.3 should now be available and should include these new languages.  Thank you again for this!

pawel-wachalski:
Hi @gregsdennis is there a chance we can get them as a separate library like es resources were published? Currently, we are using version 2.3.0. of the library and we would like to avoid update at the moment?

pawel-wachalski:
I understand the problem of repeated resources - in the main package and in the separate resource package. However if turns out troublesome, it would be helpful for me, if you publish the resources separately anyway - then it would be possible for me to achieve what I need - to have the resources while not upgrading to the newest version. How do you feel about it?

gregsdennis:
Okay... so a lot of what I tried didn't work, but it does now.

Lessons learned:

- The resources MUST be built as part of the core project.  I can't build the localizations in a separate project.
- I can't use `dotnet pack` to build because the content isn't customizable, which means the localized resources will be included.
- That means I have to have _.nuspec_ files for everything and `nuget pack` them manually.

Thanks to the Humanizer project for already tackling this and publishing their work.

gregsdennis:
The package is 1.0.1, but the lib's attributes say 4.0.0.  I've tested the packages that are uploaded to Nuget.  It all works.
[TestProject1.zip](https://github.com/gregsdennis/json-everything/files/11739059/TestProject1.zip)

The 1.0.0 versions of the lanugage packs should be unlisted.  You'll still be able to get to them, but don't use them.  I was figuring things out.  Use the 1.0.1 packages.

pawel-wachalski:
@gregsdennis in you test project you are using JsonSchema.Net 4.1.5 version (newest) and this indeed works.
In my solution I use JsonSchema.Net 2.3.0 version and when I try to use resources I get following error:
image
So the questions if if we can make the reources packages work with JsonSchema.Net 2.3.0 too?

gregsdennis:
No, it won't work with older versions. I'd have to recompile at that version.  I don't have a way to release older versions, and I'm pretty sure the messages wouldn't be compatible anyway.

gregsdennis:
I think the primary problem is that the resources DLL's `[AssemblyVersion]` attribute needs to match the main DLL's `[AssemblyVersion]` in order for .Net to load it.

I wrote a blog post (a long time ago) that touches on this: https://gregsdennis.github.io/coding-for-smarties/2016/01/21/how-to-version-assemblies-destined-for-nuget.html

pawel-wachalski:
@gregsdennis thanks for the explanations. I think there is one thing we might try, if you are willing to. In your nuspec for the resource you have defined a strict version for dependency like:
`   <dependencies>
      <dependency id="JsonSchema.Net" version="4.0.0" />
    </dependencies>`
Maybe defining it like:
``   <dependencies>
      <dependency id="JsonSchema.Net" version="2.3.0" />
    </dependencies>``
I think it might give the flexibility with the resources, not requiring to release any older versions of the main library. 

 _I'm pretty sure the messages wouldn't be compatible anyway._
 I am not sure, but there is a chance, that the resources provided, where made based on our local version of 2.3.0, not the newest - I assumed (not sure if correctly), that keys for the messages do not change overtime, maybe just a message itself.

gregsdennis:
Sorry, but the assembly versiond have to match in order for the .net runtime to load the resource assembly and associate it with the cute assembly.  Just updating the dependency in the Nuget package won't help.

PR #471, opened 2023-06-11
add support for json source generation to JsonPatch

pwelter34:
Json Source Generation requires that JsonConverter be public.  This is a small PR to make the PatchJsonConverter public to support source generation.

Reference:
https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation

### DIFF
--- a/JsonPatch.Tests/PatchExtensionTests.cs
+++ b/JsonPatch.Tests/PatchExtensionTests.cs
@@ -326,6 +326,34 @@ public class PatchExtensionTests
 		Assert.AreEqual(5, final?.Numbers?[0]);
 	}
 
+	[Test]
+	public void CreatePatch_JsonContext()
+	{
+		var initial = new TestModel
+		{
+			Id = Guid.NewGuid(),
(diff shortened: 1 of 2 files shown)


gregsdennis:
Can you explain what's going on here?  It seems this test would work just fine without that context object.  You're just creating a patch, serializing it, and verifying its serialized form.

What's being generated?

pwelter34:
The benefit to the JsonPatch object itself isn't huge.  However, when you enable source generation for all the rest of your Web API endpoint, JsonPatch also has to support source generation as it gives an error otherwise.  You have to include all objects that JsonSerializer will deal with.

The following project is an example of setting up source generation for asp.net web api.

Creating the context to enable source generation:
https://github.com/loresoft/Estimatorx/blob/JsonSerializable/src/EstimatorX.Shared/Models/JsonContext.cs

Configure asp.net web api to use the source generation:
https://github.com/loresoft/Estimatorx/blob/29e0e4740bf0fb64cf174d4c28048194d0d7988f/src/EstimatorX.Service/Program.cs#L126

As for the benefit of Json Source Generation, you can read the follow post for a good overview.
https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-source-generator/

hope that helps

~ Paul

gregsdennis:
Thanks for that.

I'll pull this into a temp branch so I can make some edits before publishing.

PR #499, opened 2023-08-04
Schema/culture option

### LINKED ISSUE #493: Support for parallel evaluation with different cultures
### Nuget Package

JsonSchema.Net

### Package Version

4.1.7

**Problem**
ErrorMessages.Culture is static property and does not support parallel evaluation of `JsonSchema` with different cultures as they can interfere each other.

**Scenario**
1. Create API endpoint where user can upload json file and select culture for error messages
2. User A sends reqeust with culture `en-us` and API starts processing request.
3. During evaluation User B sends request with culture `sv-se` and API starts processing request
4. Evaluation of User B's request override `ErrorMessages.Culture`.
5. Errors for User A are mix of English and Swedish.

**Solution**
Add culture to `EvaluationOptions`

### Describe alternatives you've considered

Currently as a workaround we are using _lock_ mechanism to force waiting until started evaluation won't be finished but it's rather poor solution regarding performance.

```
        public static EvaluationResults EvaluateWithCulture(this JsonSchema schema, JsonElement element, EvaluationOptions options = null, string culture = null)
        {
            EvaluationResults evaluationResults;
            culture ??= "en-us";

            lock (_evaluationLock)
            {
                ErrorMessages.Culture = CultureInfo.GetCultureInfo(culture);
                evaluationResults = schema.Evaluate(element, options);
                ErrorMessages.Culture = CultureInfo.GetCultureInfo("en-us");
            }

            return evaluationResults;
        }
```

_No response_

### Code of Conduct

gregsdennis:
Resolves #493

@m-adamkiewicz, this should resolve you.  I don't think there are any breaking changes as the significant stuff is internal.  Basically it adds `EvaluationOptions.Culture` that you can set instead of `ErrorMessages.Culture`.  It will fall back to the static setting when the option isn't set.

### DIFF
--- a/JsonSchema.Tests/LocalizationTests.cs
+++ b/JsonSchema.Tests/LocalizationTests.cs
@@ -1,5 +1,7 @@
-﻿using System.Globalization;
+﻿using System;
+using System.Globalization;
 using System.Text.Json.Nodes;
+using System.Threading.Tasks;
 using NUnit.Framework;
 
 namespace Json.Schema.Tests;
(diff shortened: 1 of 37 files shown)


m-adamkiewicz:
Thanks @gregsdennis, I have reviewed this PR and that's exactly what we need.
Great job and looking forward to the release.

PR #505, opened 2023-08-20
Schema/codegen

### LINKED ISSUE #504: GetKeyword method not working as expected.
### Nuget Package

JsonSchema.Net

### Package Version

5.1.3

### Operating System

Windows

.Net (5 or after)

### A clear and concise description of the bug

Seems like GetKeyword method stoped working as expected. When the keyword is not present it throws an error, but in the description of the method it states following: "Returns: The keyword if it exists; otherwise null.".

Error occures due to dictionary implementation under the hood, and it throws KeyNotFoundException exception.
Here is the error message:
"The given key 'allOf' was not present in the dictionary."

### Please add test code or steps to reproduce the behavior

Load the json schema that doesn't contain allOf keyword and call schema.GetKeyword<AllOfKeyword>(); method.

### What did you expect?

Method to return null if keyword is not present

_No response_

### Code of Conduct

### LINKED ISSUE #506: JsonSchema.FindSubschema returns null for existing path
### Nuget Package

JsonSchema.Net

### Package Version

5.1.3

### Operating System

Windows

.Net (5 or after)

### A clear and concise description of the bug

When accessing `additionalProperties` via `JsonPointer` to `((IBaseDocument)JsonSchema).FindSubschema` the call will return null.

### Please add test code or steps to reproduce the behavior

```csharp
[Test]
public void IssueXXX()
{
	IBaseDocument schema = JsonSchema.FromText("""
	{
		"type": "object",
		"additionalProperties": {
			"title": "DictTemplate",
			"type": "object",
			"properties": {
				"Enabled": {
					"type": "boolean"
				}
			}
		}
	}
	""");

	var pointer = Pointer.JsonPointer.Parse("/additionalProperties");
	var subSchema = schema.FindSubschema(pointer, EvaluationOptions.Default);

	Assert.IsNotNull(subSchema);
}
```

### What did you expect?

Since the JsonSchema is unambiguous in this case I would expect the method to return the subschema here.

The fault is at this line:
https://github.com/gregsdennis/json-everything/blob/a4aae65dd856736f42dcd0b0cf3c3bc36dfcb86b/JsonSchema/JsonSchema.cs#L521-L525

JsonPointer of "/additionalProperties" consists of only a single PointerSegment.
The extracted keyword here is a `AdditionalPropertiesKeyword` implementing `ISchemaContainer`
On pointers with more segments this will go through the switch, get unwrapped, and as the comment intended reprocessed.
With only one segment the container will go the final return and therefore be null when tried to cast to `JsonSchema`

My suggested fix, if you want I can create a PR for it:

```diff
diff --git a/JsonSchema/JsonSchema.cs b/JsonSchema/JsonSchema.cs
index ba330279..ab8d73e2 100644
--- a/JsonSchema/JsonSchema.cs
+++ b/JsonSchema/JsonSchema.cs
@@ -489,22 +489,6 @@ public class JsonSchema : IBaseDocument
 			int index;
 			switch (resolvable)
 			{
-				case ISchemaContainer container and ISchemaCollector collector:
-					if (container.Schema != null)
-					{
-						newResolvable = container.Schema;
-						i--;
-					}
-					else if (int.TryParse(segment.Value, out index) &&
-							 index >= 0 && index < collector.Schemas.Count)
-						newResolvable = collector.Schemas[index];
-
-					break;
-				case ISchemaContainer container:
-					newResolvable = container.Schema;
-					// need to reprocess the segment
-					i--;
-					break;
 				case ISchemaCollector collector:
 					if (int.TryParse(segment.Value, out index) &&
 						index >= 0 && index < collector.Schemas.Count)
@@ -519,8 +503,10 @@ public class JsonSchema : IBaseDocument
 					i += segmentsConsumed;
 					break;
 				case JsonSchema { _keywords: not null } schema:
-					schema._keywords.TryGetValue(segment.Value, out var k);
-					newResolvable = k;
+					schema._keywords.TryGetValue(segment.Value, out var keyword);
+					newResolvable = keyword is ISchemaContainer { Schema: not null } kcontainer
+						? kcontainer.Schema
+						: keyword;
 					break;
 			}
 
```

Alternative:

```diff
diff --git a/JsonSchema/JsonSchema.cs b/JsonSchema/JsonSchema.cs
index ba330279..9513b5d3 100644
--- a/JsonSchema/JsonSchema.cs
+++ b/JsonSchema/JsonSchema.cs
@@ -538,7 +538,12 @@ public class JsonSchema : IBaseDocument
 			resolvable = newResolvable!;
 		}
 
-		return resolvable as JsonSchema;
+		return resolvable switch
+		{
+			JsonSchema schema => schema,
+			ISchemaContainer container => container.Schema,
+			_ => null
+		};
 	}
 
 	/// <summary>
```

### Code of Conduct

gregsdennis:
Very basic.  Supports

- numbers (`int` & `double`)
- booleans
- strings
- arrays (`List<T>` for named types)
- dictionaries (currently only string-keyed)
- custom objects

`$ref` is generally handled, but there may be some cases that aren't.

---

Also resolves #504
Also resolved #506


gregsdennis:
Converting this to draft as I don't want to accidentally merge it.  In writing the docs, it seems that there's still quite a bit of functionality that should probably go into an initial release, even one at 0.1.0.

I'm going to work on the other pieces of this effort locally and probably release multiple packages simultaneously.

- `dotnet tool` CLI
- source generator
- OpenAPI support via [Graeae](https://github.com/gregsdennis/Graeae)

PR #511, opened 2023-08-30
fix: make sure the sub schema cache is thread-safe for JsonNodeBaseDocument

Fresa:
We've seen concurrency problems in `JsonNodeBaseDocument` that this should fix.
Example:
```
System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
     at System.Collections.Generic.Dictionary`2.FindValue(TKey key)
     at System.Collections.Generic.Dictionary`2.TryGetValue(TKey key, TValue& value)
     at Json.Schema.JsonNodeBaseDocument.FindSubschema(JsonPointer pointer, EvaluationOptions options)
     at Json.Schema.RefKeyword.Evaluate(EvaluationContext context)
     at Json.Schema.EvaluationContext.Evaluate()
```

### DIFF
--- a/JsonSchema/JsonNodeBaseDocument.cs
+++ b/JsonSchema/JsonNodeBaseDocument.cs
@@ -49,11 +49,9 @@ public class JsonNodeBaseDocument : IBaseDocument
 	{
 		return _foundSubschemas.GetOrAdd(pointer, jsonPointer =>
 		{
-			if (_foundSubschemas.TryGetValue(jsonPointer, out var schema)) return schema;
-
 			if (!jsonPointer.TryEvaluate(_node, out var location)) return null;
 
-			schema = location.Deserialize<JsonSchema>();
+			var schema = location.Deserialize<JsonSchema>();
 			if (schema != null)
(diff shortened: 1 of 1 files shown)


gregsdennis:
Thanks for this.  The change looks good.

I'm going to change the target branch so that I can make some release edits, then I'll pull it into `master`.

gregsdennis:
@Fresa would you be able to create some tests (at least one) for this, please?  I'll run them locally to see that they fail without the change.

Fresa:
@gregsdennis I don't see how tests could be written for this as it's an uncommon concurrency issue which should be solved by this fix. I churned some tests in parallel using this functionality yesterday for a good amount of time and couldn't reproduce it.

Do you have any suggestions?

gregsdennis:
Hmm.  I do have a test that runs parallel checks for another case.  I'd have thought that approach world work here too.  Let me try a bit tomorrow.

gregsdennis:
I, too, have been unable to break it intentionally.  I'll go ahead and pull this in because it looks like it should do the thing.

Thanks.

PR #519, opened 2023-09-05
JsonSchema: include line numbers and byte positions in JsonException

### LINKED ISSUE #517: JsonException line numbers do not indicate where issue occurred when deserializing JsonSchema
### Nuget Package

Json.More.Net

### Package Version

5.2.0

### Operating System

None

None

### A clear and concise description of the bug

When deserializing Jsonschema, line numbers always indicate the start of the JsonSchema.

### Please add test code or steps to reproduce the behavior

See https://github.com/mdekrey/json-everything/blob/test/json-line-numbers/JsonSchema.Tests/GithubTests.cs#L880-L894
```
	[Test]
	public void IssueTBD_IncorrectJsonExceptionpointer()
	{

		var schemaStr = @"{
  ""$schema"": ""https://json-schema.org/draft/2020-12/schema"",
  ""required"": [ { ""abc"": null } ]
  ""properties"": {
      ""abc"": { ""$ref"": ""mySchema.json"" }
  },
  ""additionalProperties"": false
}";
		var exception = Assert.Throws<JsonException>(() => JsonSerializer.Deserialize<JsonSchema>(schemaStr));
		Assert.IsNotNull(exception);
		Assert.AreEqual(2L, exception?.LineNumber);
	}
```

### What did you expect?

`LineNumber` and `BytePositionInLine` should correspond to the position of the error.

According to https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/converters-how-to?pivots=dotnet-7-0, the correct way to write a `JsonConverter` uses `JsonSerializerOptions.GetConverter` to get appropriate converter types. By using `JsonSerializer.Deserialize`, the line numbers are stripped.

Converting to `JsonSerializerOptions.GetConverter` solves this issue.

I'm able to provide a PR to resolve this issue.

### Code of Conduct

mdekrey:
Updates Json.Schema to use `JsonSerializerOptions.GetConverter`. When a generic instance is not available (such as when getting converters via Keyword types), uses an `ArbitraryDeserializer` internal class with an interface.

Intended remaining tasks:
- [X] More tests to check ~all~ relevant modified keywords
- [x] Consider using an `Instance` member on `ArbitraryDeserializer` to reduce garbage collection
- [x] Capture benchmark before and after

Resolves #517


JsonSchema.Tests/GithubTests.cs
```diff
@@ -875,4 +875,25 @@ public void Issue506_ReffingIntoAdditionalProperties()
...
+	[TestCase(@"{""anyOf"":[""not-a-schema""]}", 0, 24)]
+	[TestCase(@"{
+  ""$schema"": ""https://json-schema.org/draft/2020-12/schema"",
+  ""required"": [ { ""abc"": null } ]
+  ""properties"": {
+      ""abc"": { ""$ref"": ""mySchema.json"" }
+  },
+  ""additionalProperties"": false
+}", 2, 17)]
+	public void Issue517_IncorrectJsonExceptionpointer(string schemaStr, long expectedLineNumber, long expectedBytePositionInLine)
```
gregsdennis, line 891: Can you rename this, please? "Pointer" suggests JSON Pointer.

Also, are the values in the exception `long`s?  The expected values' types should match.

Json.More/EnumStringConverter.cs
```diff
@@ -94,7 +94,7 @@ public override T Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerial
 					values.Add(immediate);
 					reader.Read();
 				}
-				reader.Read(); // waste the end array
+				// end of the array should be left unread - the first thing the parent converter will do is call Read() again
```
gregsdennis, line 97: 👍 nice catch. no need for the comment, though.

gregsdennis (commented):
Looking good so far.  I like the direction.

gregsdennis:
> More tests to check all modified keywords

I typically rely on the JSON Schema test suite pretty heavily to make sure everything's working.  I think some tests for a few different keywords of different shapes (e.g. `allOf` vs `not` vs `properties`, etc.) should do for validating they error properly.

JsonSchema.Tests/GithubTests.cs
mdekrey: I gave `long` because that's what's in `JsonException`, but all of the examples I'm writing are going to be short enough that it won't need a `long`. I could go either way with the type for the test.

gregsdennis:
I've rebased this onto #518 so that everything can be released at once.  (I used GitHub, so let's hope everything still compiles 😆)

JsonSchema.Tests/GithubTests.cs
gregsdennis: I didn't have time to check what was in the exception at the time.  Let's use what's there.  (Though I'm surprised they used a long.  Do they expect lines longer than 2 billion chars?)

JsonSchema/JsonSerializerOptionsExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Collections.Generic;
+using System.Text;
+using System.Text.Json;
+using System.Text.Json.Nodes;
+using System.Text.Json.Serialization;
+
+namespace Json.Schema;
+
+// TODO: .NET 5+ would have these methods marked with `RequiresUnreferencedCodeAttribute` to warn against tree trimming
+internal static class JsonSerializerOptionsExtensions
```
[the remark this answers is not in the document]
gregsdennis: I created Json.More specifically to hold more general-use extensions to STJ that I think probably should have been included.  This seems to fit.

I'm happy with leaving the `object? Read()` method here since it does seem to solve this specific problem of not knowing what we're going to get back.

mdekrey:
> Consider using an Instance member on ArbitraryDeserializer to reduce garbage collection

This didn't seem to cause any differences in benchmark results. 

> Capture benchmark before and after

Benchmarks in related issue, at https://github.com/gregsdennis/json-everything/issues/517#issuecomment-1706514417

JsonSchema/JsonSerializerOptionsExtensions.cs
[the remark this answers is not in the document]
mdekrey: I already marked this ready for review, but I'm resolving this comment in a commit right now, sorry for the early notification.

Json.More/EnumStringConverter.cs
```diff
@@ -94,7 +94,6 @@ public override T Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerial
 					values.Add(immediate);
 					reader.Read();
 				}
-				reader.Read(); // waste the end array
```
gregsdennis, line 97: Actually, this is needed.  It's handling flags enums, which render as arrays.  It mirrors consuming the array start.

The entire array needs to be consumed as part of this.  If there's an error, it's being caused somewhere else.

gregsdennis: ... unless the read on 95 does it?  Probably a good idea to have some tests around this (which I didn't appear to do).

JsonSchema/JsonSerializerOptionsExtensions.cs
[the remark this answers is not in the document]
gregsdennis: Generally, I prefer the commenter to resolve their own comment to assure that their concern has been addressed.

JsonSchema/RequiredKeyword.cs
```diff
@@ -82,13 +82,7 @@ internal class RequiredKeywordJsonConverter : JsonConverter<RequiredKeyword>
...
 	public override RequiredKeyword Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
 	{
-		using var document = JsonDocument.ParseValue(ref reader);
-
-		if (document.RootElement.ValueKind != JsonValueKind.Array)
-			throw new JsonException("Expected array");
-
-		return new RequiredKeyword(document.RootElement.EnumerateArray()
-			.Select(e => e.GetString()!));
+		return new RequiredKeyword(options.Read<IEnumerable<string>>(ref reader) ?? throw new JsonException("Expected array"));
```
gregsdennis, line 85: Is there a reason that you're using `IEnumerable` here instead of just an array?

JsonSchema/Serialization/ValidatingJsonConverter.cs
```diff
@@ -112,6 +112,7 @@ public ValidatingJsonConverter(JsonSchema schema, Func<JsonSerializerOptions, Js
 		var newOptions = _optionsFactory(options);
 
 		if (validation.IsValid)
+			// TODO: this does not use the options reader because at least one test failed with a NotSupportedException
```
gregsdennis, line 115: I'd like this explored, please.

JsonSchema/RequiredKeyword.cs
mdekrey: The `.Select` previously returned an `IEnumerable`; I matched previous types. However, the JsonSerializerOptions this way will determine the type; the default as determined by the .NET team to hopefully be the most performant option, which for now is probably still an array.

Json.More/EnumStringConverter.cs
mdekrey: Hmm, I'll go check that, though it definitely seemed like an error. I'd be happier if we didn't need to change that, so I'll go verify!

mdekrey: Yeah, I'm not sure what led me to that line in the first place anymore; replacing it changed no test results. Thanks for calling it out!

JsonSchema/RequiredKeyword.cs
gregsdennis: > which for now is probably still an array.

... which is then put into a list in the constructor.

Json.More/EnumStringConverter.cs
mdekrey: I'm seeing an additional 42 tests failing with this added back, many of them "Expected keyword or end of schema object", such as through SelfValidationTests.

Per https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/converters-how-to?pivots=dotnet-7-0#the-use-of-utf8jsonreader-in-the-read-method:
> If you read beyond the end of the object, or if you stop before reaching the corresponding end token, you get a JsonException exception indicating that:
> 
>> The converter 'ConverterName' read too much or not enough.
While I'm not seeing that specific errors, I'm seeing something similar, such as on `RoundTrip("{\"type\":[\"object\",\"string\"]}")`, it reports "Expected token" from the `throw` in `SchemaJsonConverter.Read`.

JsonSchema/RequiredKeyword.cs
mdekrey: I definitely didn't dig in that far! I'll switch that to an array so it uses the `params` constructor.

JsonSchema/Serialization/ValidatingJsonConverter.cs
mdekrey: The specific message:

> System.NotSupportedException: 'Metadata for type 'Json.Schema.Tests.Serialization.DeserializationTests+FooWithSchema' was not provided to the serializer. The serializer method used does not support reflection-based creation of serialization-related type metadata. If using source generation, ensure that all root types passed to the serializer have been indicated with 'JsonSerializableAttribute', along with any types that might be serialized polymorphically.'

It sounds like the source generators aren't picking up the type transitively; I'm not familiar enough with how those work to solve it at this time.

Json.More/JsonSerializerOptionsExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+
+	private class ArbitraryDeserializer<T> : IArbitraryDeserializer
+	{
+		public object? Read(ref Utf8JsonReader reader, JsonSerializerOptions options)
+		{
+			var converter = (JsonConverter<T>)options.GetConverter(typeof(T));
+
+			return converter.Read(ref reader, typeof(T), options);
+		}
+	}
```
gregsdennis, line 66: I agree that these can be cached.  Looking again, I think it makes sense to do it here instead of another PR.

I think something like this would do:

- Change `IArbitraryDeserializer` to abstract class `ArbitraryDeserializer`
- Add private static `ConcurrentDictionary<Type, JsonSchema>` as a cache to abstract base
- Add static `GetConverter()` to abstract base

Json.More/JsonSerializerOptionsExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+/// </summary>
+public static class JsonSerializerOptionsExtensions
+{
+	/// <summary>
+	/// Returns the converter for the specified type.
+	/// </summary>
+	/// <typeparam name="T">The <see cref="Type"/> to convert.</typeparam>
+	/// <param name="options">The <see cref="JsonSerializerOptions"/> being used.</param>
+	/// <returns>An implementation of <see cref="JsonConverter<>"/> as determined by the provided options</returns>
+	public static JsonConverter<T> GetConverter<T>(this JsonSerializerOptions options)
```
gregsdennis, line 19: Is this used outside of this class?  I don't see it.  If not, it can be internal.

Json.More/JsonSerializerOptionsExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	/// <summary>
+	/// Read and convert the JSON to an arbitrary type.
+	/// </summary>
+	/// <param name="options">The <see cref="JsonSerializerOptions"/> being used.</param>
+	/// <param name="reader">The <see cref="Utf8JsonReader"/> to read from.</param>
+	/// <param name="arbitraryType">The <see cref="Type"/> to convert to.</param>
+	/// <returns>The value that was converted.</returns>
+	public static object? Read(this JsonSerializerOptions options, ref Utf8JsonReader reader, Type arbitraryType)
```
gregsdennis, line 46: I think this one can stay in JsonSchema project (and internal).  I don't know that it's as generally useful.

Json.More/EnumStringConverter.cs
gregsdennis: Hm... that would likely be on the `type` keyword (the only place I'm aware that uses flags enums in the serialization).  I'll download the merge branch and poke around a bit.

mdekrey: I force-pushed to remove the commit where I removed this line; adding it in did not resolve any tests. Please let me know if there's something else I can look into for this.

I believe existing tests cover this scenario, but were not catching the flaw previously because the converter was being used directly thanks to the extra `JsonSerializer.Deserialize` calls; nothing was using the `Utf8JsonReader` after it had the extra Read.

gregsdennis: Sounds good.

I've also just found that the converter serializes the zero-value name if there is one 🤦.  I'll deal with that later.

Json.More/JsonSerializerOptionsExtensions.cs
mdekrey: Yeah, though it could be useful; there are plenty of reasons to use the converter.

For instance, the `HandleNull` property is supposed to be used to determine whether or not to write out `null`. This may be the expected option to avoid the `SchemaKeywordRegistry.GetNullValuedKeyword` situation - I'm not exactly sure what caused the need for that!

gregsdennis: That actually looks like it might legacy code that isn't used anymore.  I'd have to dig into older versions to figure out why I needed it.  I might still be needed and that it doesn't use it may be broken.

Json.More/JsonSerializerOptionsExtensions.cs
mdekrey: Couldn't add the `GetConverter()` to the abstract base since it doesn't have the generic at that level, but I did add the cache and use an abstract base class as suggested.

gregsdennis: So the point of making the base type abstract is to house the cache in it.  Then the `.GetConverter()` method is defined on that.  It encapsulates everything into the deserializer.  Maybe that's overengineering it.

I think what you've done is fine, and the abstract base class can go back to an interface.

JsonSchema/Serialization/ValidatingJsonConverter.cs
gregsdennis: Okay.  I'll explore this a bit.

Json.More/JsonSerializerOptionsExtensions.cs
mdekrey: Ah, I misunderstood; got you now! Added one more commit for this - it does make it cleaner. Thanks for the advice!

gregsdennis:
This looks good to me.  I'll pull it into that other branch and make some personal tweaks (and try to figure out that flag enum thing).  Thanks for this work!

PR #528, opened 2023-09-21
added ability to use a custom name resolving logic PropertyNameResolvingMethods so users can provide their custom …

juwens:
The current code provides only a very limited possibility to create the property-name via a custom logic. Namely with PropertyNamingMethod.

For our case this wasn't enough. We wanted to extract the name from ConfigurationKeyNameAttribute.
But currently, the only property which can be used is JsonPropertyName, and it is hardcoded in ObjectSchemaGenerator.cs.

This PR enables the user to provide a custom logic to resolve a property-name with the full information the MemberInfo can provide. Not only the `member.Name` of PropertyNamingMethod.

example for defining a custom logic, analogue to PropertyNamingMethod.

```c#
new SchemaGeneratorConfiguration()
{
    PropertyNameResolvingMethod = member => member.GetCustomAttributes<ConfigurationKeyNameAttribute>().FirstOrDefault()?.Name;
}
```

Background (why do we need it):

We are creating a schema for our .net configuration, hence using the ConfigurationKeyName is a necessity.
The workaround would be to add ConfigurationKeyName AND JsonPropertyName to every property, but that  would be quite ugly.

FTR: https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.configurationkeynameattribute?view=dotnet-plat-ext-7.0


JsonSchema.Generation/PropertyNameResolvingMethods.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+	/// <summary>
+	/// Property name is read from <see cref="JsonPropertyNameAttribute"/>.
+	/// </summary>
+	public static readonly PropertyNameResolvingMethod ByJsonPropertyName = member => member.GetCustomAttributes<JsonPropertyNameAttribute>().FirstOrDefault()?.Name;
+
+	/// <summary>
+	/// Don't do any resolving. Not even <see cref="JsonPropertyNameAttribute"/> will be read.
+	/// </summary>
+	public static readonly PropertyNameResolvingMethod None = _ => null;
```
gregsdennis, line 27: What's the purpose of this?  It's only going to cause exceptions.

JsonSchema.Generation/PropertyNameResolvingMethods.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Text.Json.Serialization;
+
+namespace Json.Schema.Generation;
+
+/// <summary>
+/// Declares a property name resolution which is used to provide a property name.
+/// </summary>
+/// <param name="input">The property.</param>
+/// <returns>The property name or null, if no name was resolved.</returns>
+public delegate string? PropertyNameResolvingMethod(MemberInfo input);
```
gregsdennis, line 12: Please rename to `PropertyNameResolver` (and the config property as well).

gregsdennis (changes requested):
Overall, I think I'd perfer to have a single naming mechanism. Can you deprecate `PropertyNamingMethod` (add `[Obsolete]` with a message) and forward its implementation to the new config property?

JsonSchema.Generation/PropertyNameResolvingMethods.cs
juwens: i removed it.

JsonSchema.Generation/PropertyNameResolvingMethods.cs
juwens: done

juwens:
BTW: very nice project you made here.

> Thanks for the idea. I didn't realize this might be a need.

I guess it's not very common. Our config is just so big and potentially changing (subtle or complex) with new releases, that we wanted to make things easier/comprehensible for our customer.

> Overall, I think I'd perfer to have a single naming mechanism. Can you deprecate `PropertyNamingMethod` (add `[Obsolete]` with a message) and forward its implementation to the new config property?

Done the obsolete think, and went the extra mile of unifying both systems. The PropertyNameResolver property in the config now is the single source of truth for deriving a name. So that after an official deprecation the old property can simply be removed.

The code now consumes only the new PropertyNameResolver. (`var name = SchemaGeneratorConfiguration.Current.PropertyNameResolver(member);`)

image

Not the `var name = resolver() ?? naming();` thing it did in the initial commit.

The whole forward/backward compat "magic" is here: https://github.com/gregsdennis/json-everything/blob/b399f3bd5ec2b54835b369ca26f6c1d40663071e/JsonSchema.Generation/SchemaGeneratorConfiguration.cs#L53

I added tests for the PropertyNameResolver, similar to the old ones.

JsonSchema.Generation/Generators/ObjectSchemaGenerator.cs
```diff
@@ -184,24 +181,24 @@ private static IEnumerable<IConditionAttribute> ExpandEnumConditions(IConditionA
 				var values = Enum.GetValues(memberType);
 				foreach (var value in values)
 				{
-					yield return new IfAttribute(ifEnumAttribute.PropertyName, ifEnumAttribute.UseNumbers ? value : value.ToString(), value);
+					yield return (new IfAttribute(ifEnumAttribute.PropertyName, ifEnumAttribute.UseNumbers ? value : value.ToString(), value), member);
```
juwens, line 184: it's a technical consequence of the unification.

But, are you fine with doing it? I didn't dive into the whole condition thing yet, just modified it _to the best of my belief_.

JsonSchema.Generation/Generators/ObjectSchemaGenerator.cs
```diff
@@ -68,10 +68,7 @@ public void AddConstraints(SchemaGenerationContextBase context)
 
 			var memberContext = SchemaGenerationContextCache.Get(member.GetMemberType(), unconditionalAttributes);
 
-			var name = SchemaGeneratorConfiguration.Current.PropertyNamingMethod(member.Name);
-			var nameAttribute = unconditionalAttributes.OfType<JsonPropertyNameAttribute>().FirstOrDefault();
-			if (nameAttribute != null)
-				name = nameAttribute.Name;
+			var name = SchemaGeneratorConfiguration.Current.PropertyNameResolver(member);
```
juwens, line 71: This the most significant change. It must be seen together with the implementation of the consumed property.

JsonSchema.Generation/SchemaGeneratorConfiguration.cs
```diff
@@ -34,11 +35,42 @@ public class SchemaGeneratorConfiguration
...
+
+	/// <summary>
+	/// Gets or sets the property name resolving method. Default is <see cref="PropertyNameResolvers.ByJsonPropertyName"/>.
+	/// </summary>
+	/// <remarks>
+	/// This can be replaced with any `Func&lt;MemberInfo, string&gt;`.
+	/// </remarks>
+	public PropertyNameResolver PropertyNameResolver
+	{
+		get
```
juwens, line 53: You might not be 100% happy with the complex getter. Cause the rest of the class is just a relatively simple Poco.

But, the logic for the forward/backward compat needed to go somewhere, and all of the if-return can be removed, when the old prop will be removed.

juwens:
i added three comments in the review to point to noteworthy aspects.

JsonSchema.Generation/SchemaGeneratorConfiguration.cs
```diff
@@ -34,11 +35,42 @@ public class SchemaGeneratorConfiguration
...
+			}
+
+			// backwards compatibility: if the old property was actively set in existing code existing code, code will use it
+			if (PropertyNamingMethod is PropertyNamingMethod method)
+			{
+				return new PropertyNameResolver(x => method(x.Name));
+			}
+
+			// if neither property was actively set, the behavior is as it was before. Prefer JsonPropertyName and Fallback to AsDeclared
+			return PropertyNameResolvers.ByJsonPropertyName;
```
gregsdennis, line 68: The attribute should override this setting completely.  If someone has explicitly stated the name they want to use in the JSON, then the generator needs to use that, regardless of what this setting is.  This shouldn't be part of the setting.

The default here needs to be `AsDeclared`, like it is currently for `PropertyNamingMethod`.

JsonSchema.Generation/SchemaGeneratorConfiguration.cs
```diff
@@ -34,11 +35,42 @@ public class SchemaGeneratorConfiguration
...
 	/// This can be replaced with any `Func&lt;string, string&gt;`.
 	/// </remarks>
-	public PropertyNamingMethod PropertyNamingMethod
+	[Obsolete($"Use {nameof(PropertyNameResolver)} instead.")]
+	public PropertyNamingMethod? PropertyNamingMethod
 	{
-		get => _propertyNamingMethod ??= PropertyNamingMethods.AsDeclared;
+		get => _propertyNamingMethod;
 		set => _propertyNamingMethod = value;
 	}
```
gregsdennis, line 43: Instead of all of the conditional logic in `PropertyNameResolver`, this should actually be implemented _using_ `PropertyNameResolver`.  Something like this:

```c#
[Obsolete($"Use {nameof(PropertyNameResolver)} instead.")]
public PropertyNamingMethod? PropertyNamingMethod
{
	get => x => PropertyNameResolver(new DummyInfo(x));
	set => PropertyNameResolver = x => x.Name;
}

private class DummyInfo : MemberInfo
{
	public override object[] GetCustomAttributes(bool inherit) => Array.Empty<object>();

	public override object[] GetCustomAttributes(Type attributeType, bool inherit) => Array.Empty<object>();

	public override bool IsDefined(Type attributeType, bool inherit) => false;

	public override Type DeclaringType { get; } = typeof(DummyInfo);
	public override MemberTypes MemberType { get; } = MemberTypes.Property;
	public override string Name { get; }
	public override Type? ReflectedType { get; } = null;

	public DummyInfo(string name)
	{
		Name = name;
	}
}
```

Then `PropertyNameResolver` can be just an autoproperty.

Since this library isn't calling `PropertyNamingMethod` anymore, its getter _should_ never be called, so we don't need to worry about it instantiating a bunch of objects.

JsonSchema.Generation/assembly.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("JsonSchema.Generation.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100d111b5991feb360746cf4535b9ec3196825996680b6599cf10748f90a58ec820431f81d21eff5229a3fd7c3e42e46a20584ac413ebe39b4c05905fd47b877bffed36bff23ad4ad14aad3deed8f24ed40a1371daad79a41172b2cdf0db4b664198f38f3a19419457131f8ee4558e628b5c986bb574d4e151b650eae6aa1f5ccbf")]
```
gregsdennis, line 3: Why?

JsonSchema.Generation/Generators/ObjectSchemaGenerator.cs
```diff
@@ -68,10 +68,7 @@ public void AddConstraints(SchemaGenerationContextBase context)
 
 			var memberContext = SchemaGenerationContextCache.Get(member.GetMemberType(), unconditionalAttributes);
 
-			var name = SchemaGeneratorConfiguration.Current.PropertyNamingMethod(member.Name);
-			var nameAttribute = unconditionalAttributes.OfType<JsonPropertyNameAttribute>().FirstOrDefault();
-			if (nameAttribute != null)
-				name = nameAttribute.Name;
```
gregsdennis, line 74: As stated before, this is the preferred behavior.  `[JsonPropertyName]` should override this setting, not be a part of it.

JsonSchema.Generation/PropertyNameResolvers.cs
```diff
@@ -0,0 +1,75 @@
...
+	/// </summary>
+	public static readonly PropertyNameResolver KebabCase = x => x.Name.Kebaberize();
+	/// <summary>
+	/// Property names to upper kebab case (e.g. `UPPER-KEBAB-CASE`).
+	/// </summary>
+	public static readonly PropertyNameResolver UpperKebabCase = x => x.Name.Kebaberize().ToUpperInvariant();
+	/// <summary>
+	/// Property name is read from <see cref="JsonPropertyNameAttribute"/>, falls back to <see cref="AsDeclared"/> if attribute is absent.
+	/// </summary>
+	public static readonly PropertyNameResolver ByJsonPropertyName = x => PartialPropertyNameResolvers.ByJsonPropertyName(x) ?? AsDeclared(x);
```
gregsdennis, line 63: Remove.  Then all of the "partial" stuff can be removed, too.

gregsdennis (changes requested):

JsonSchema.Generation/SchemaGeneratorConfiguration.cs
juwens: Actually, the way i implemented it is exactly the current behavior:

In the current main it reads the name via PropertyNamingMethod which has AsDeclard as default.
And then if there is a JsonPropertyNameAttribute, it overrides the name with it.

Which is what I replaces a more flexible/mix-and-match/lego-brick way.

image

juwens: I mean, you'll always need a fallback, it an Attribute is not present.

JsonSchema.Generation/SchemaGeneratorConfiguration.cs
juwens: done that, it's now a simple autoproperty

JsonSchema.Generation/PropertyNameResolvers.cs
juwens: done

JsonSchema.Generation/Generators/ObjectSchemaGenerator.cs
juwens: I tried to implement it the way you wanted.

JsonSchema.Generation/assembly.cs
juwens: removed, now that there is no internal class anymore

juwens:
I tried to understand your requirements, and commited according changes, but i don't see, how your suggestions combined with backward compatibility with the master can work.

Feel free to directly apply your suggestions, i have the feeling, it might be easier if you add the few lines the way you like.

JsonSchema.Generation/SchemaGeneratorConfiguration.cs
gregsdennis: Yes, but the intentionality of the code is that `[JsonPropertyName]` trumps everything, regardless of the setting.  This is the behavior I want, and the code should read the same.

JsonSchema.Generation/PropertyNameResolvers.cs
```diff
@@ -0,0 +1,56 @@
...
+	/// </summary>
+	public static readonly PropertyNameResolver KebabCase = x => x.Name.Kebaberize();
+	/// <summary>
+	/// Property names to upper kebab case (e.g. `UPPER-KEBAB-CASE`).
+	/// </summary>
+	public static readonly PropertyNameResolver UpperKebabCase = x => x.Name.Kebaberize().ToUpperInvariant();
+	/// <summary>
+	/// Property name is read from <see cref="JsonPropertyNameAttribute"/>.
+	/// </summary>
+	public static readonly PropertyNameResolver ByJsonPropertyName = x => x.GetCustomAttributes<JsonPropertyNameAttribute>().FirstOrDefault()?.Name /* TODO: what do we do with the nullable here? */;
```
gregsdennis, line 55: This shouldn't be a resolver.  Please restore the [`if` statement](https://github.com/gregsdennis/json-everything/pull/528/files#r1334900430) and remove this.

Issue #552, opened 2023-11-03
Transforming Rule to Expression Tree for LINQ?


cdroulers:
### Nuget Package

JsonLogic

### Package Version

latest

### How can I help?  Please provide as much context as possible.

Hello!

I'm trying to find an easy, portable way to represent user-entered filter parameters to a database. The filters will be fairly simple, but I always like to find something a little standardised to avoid re-inventing the wheel. I've looked at many custom JSON schemas, CEL, SpEL, etc.

After finding https://react-querybuilder.js.org/ for the eventual UI (won't be using it right away), I discovered JSON logic and this library.

I'm currently trying to figure out if there's an existing way to transform JSON logic to a .NET Expression tree that can be passed to EFCore, ideally with possibilty for constants to parameterized, but that can be fixed later.

I see that all properties are internal to avoid mutability, but having them exposed publicly (through immutable collections) could allow me to visit each node and transform to expression tree.

Or maybe there is already a visitor pattern that can be applied, but I couldn't really figure that out from the very little docs I could find! Everything being internal prevents navigating the tree of rules (as mentioned in this issue https://github.com/gregsdennis/json-everything/issues/154)

I'm not fully set on using JsonLogic, but I'm throwing the question out here in case there's a way I'm missing or some other way that could make sense! 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gregsdennis:
I currently don't have such functionality, no.  That kind of translation will take some planning.  If you have some ideas, I'd be happy to hear them.

cdroulers:
Hi Greg! Thanks for the answer!

We're looking to store some filter logic in our database in a way that's easy to understand for humans and machines alike. We use the filter to query our DB and set some permissions according to query results.

I've written a small POC over the weekend using SQL

```csharp
where = GetWhere("Name = 'TEST' OR Id IN (1, 2)");
var exp = GetWhereExpressionLambda<TestData>(where.SearchCondition);
```
Walking the SQL structure (using `Microsoft.SqlServer.TransactSql.ScriptDom`) I can generate an Expression tree (using Expression.Lambda and other methods)

```
// {x => ((x.Name == "TEST") Or value(System.Collections.Generic.List`1[System.Int32]).Contains(x.Id))}
.Lambda #Lambda1<System.Func`2[ProjectName.UnitTest1+TestData,System.Boolean]>(ProjectName.UnitTest1+TestData $x)
{
[...]

PR #562, opened 2023-11-12
[WIP] Add basic POC to build Expression tree for and/or, variable and literal rules.

cdroulers:
@gregsdennis Here's a quick POC I made late at night, related to issue https://github.com/gregsdennis/json-everything/issues/552 Get the basics across, the best way to actually do that would probably be a visitor, to reduce the code inside each rule and avoid passing the ParameterExpression in all the nodes.

I made this for fun, but it could be practical for someone else!

### DIFF
--- a/JsonLogic.Tests/StrictEqualsTests.cs
+++ b/JsonLogic.Tests/StrictEqualsTests.cs
@@ -1,3 +1,6 @@
+using System;
+using System.Linq;
+using System.Linq.Expressions;
 using Json.Logic.Rules;
 using NUnit.Framework;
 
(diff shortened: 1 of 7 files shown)


gregsdennis:
This looks promising!  At first I was wondering how the other JSON Logic operators would fit into this, but then I realized that it's not up to me (as a tool maker).  It's up to the users to figure out how they want to use it.

I'm working on implementing JSON-e right now, but I can probably take this on next.

cdroulers:
As I mentioned, I have no plans on using something like this anymore, I've settled on using SQL as the filtering language and parsing the WHERE condition similarly to this POC. So absolutely no need to spend time on this, I just wanted to throw it out there and see if it could be useful to anybody else!

As for operators, most look like they would easily translate to Expressions, except maybe `map`, `reduce` and `filter`.
Although `map` could be translated to a `.Select()` projection and `reduce` to `.Aggregate()`.

I don't know if you'd like to keep this PR open, but feel free to close it to remove clutter!

PR #564, opened 2023-11-17
update json converters to support source gen

### LINKED ISSUE #549: Is it possible to make custom JSON converters public to support source generation?
### Describe the functionality that isn't covered by this suite.

I would like to utilize source generation in `System.Text.Json` for some POCOs that rely on types from this package, such as `JsonPointer`. However, it appears that the source generator consistently includes the custom JsonPointerJsonConverter, which is `internal` and is applied to the `JsonPointer` class, leading to a build failure.

The issue can be reproduced with the following code:

```C#
// Build error: 'JsonPointerJsonConverter' is inaccessible due to its protection level
public record ErrorResponse(string Code, string Message, JsonPointer? Target);

[JsonSerializable(typeof(ErrorResponse))]
public partial class ErrorResponseContext : JsonSerializerContext { }
```

Would it be possible for you to consider changing the visibility of the custom JSON converters to `public`?

### Is there any other information you'd like to share regarding this functionality?

_No response_

### Code of Conduct

gregsdennis:
Resolves #549

### DIFF
--- a/JsonPointer/JsonPointer.csproj
+++ b/JsonPointer/JsonPointer.csproj
@@ -4,9 +4,9 @@
     <TargetFramework>netstandard2.0</TargetFramework>
     <LangVersion>latest</LangVersion>
     <RootNamespace>Json.Pointer</RootNamespace>
-    <Version>3.1.0</Version>
+    <Version>3.2.0</Version>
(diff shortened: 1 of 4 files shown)


gregsdennis:
I'll follow this up with updates for the other libs.

PR #566, opened 2023-11-24
JSON-e

### LINKED ISSUE #539: JSON-e
### Describe the functionality that isn't covered by this suite.

JSON-compatible templating/transforming syntax.

https://json-e.js.org/using.html
https://github.com/json-e/json-e

JSONata isn't JSON-compatible.

### Is there any other information you'd like to share regarding this functionality?

_No response_

### Code of Conduct

gregsdennis:
Resolves #539
Depends on https://github.com/json-e/json-e/pull/504

There are three tests that involve unicode that I believe are incorrect as they're not properly JSON-escaping the chars.  The dependency above is tracking that correction.

Additionally, while I strived to make error messages the same, they're not on about 30 tests.  The scenarios still produce errors, as expected, but the message isn't the same.  I'm considering this good enough for a first release.


gregsdennis:
Going to release as is for now, ignoring the dependency.  When that goes through, I'll un-ignore those tests and push an update if necessary.

Issue #570, opened 2023-12-01
`time-offset` should be required for `date-time`


danielstarck:
### Nuget Package

JsonSchema.Net

### Package Version

5.4.0

### Operating System

Linux

### .Net Target (if relevant, please specify the version in the description)

.Net (5 or after)

### A clear and concise description of the bug.

Instances like `"2023-01-01T12:34:56"` (essentially `date-time`s with the `time-offset` part omitted) are considered valid `date-time`s.

### What did you expect?

I expected `IsValid` to be `false`.

### Please add test code or steps to reproduce the behavior.

```
[Fact]
public void Evaluate_WhenGivenDateTimeWithMissingTimeOffset_ShouldReturnIsValidFalse()
{
    const string schemaString =
        """
        {
          "$schema": "https://json-schema.org/draft/2020-12/schema",
          "type": "object",
          "properties": {
            "missingTimeOffset": {
              "format": "date-time"
            }
          }
        }
        """;

    const string json =
        """
        {
          "missingTimeOffset": "2023-01-01T12:34:56"
        }
        """;

    var jsonDocument = JsonDocument.Parse(json);
    var evaluationOptions = new EvaluationOptions { RequireFormatValidation = true };
    var isValid = JsonSchema.FromText(schemaString).Evaluate(jsonDocument, evaluationOptions).IsValid;

    Assert.False(isValid);
}
```

### Is there any other information you'd like to share regarding this bug?

I can't find anything in [draft 2020-12](https://json-schema.org/draft/2020-12/json-schema-validation#name-dates-times-and-duration) (or the referenced [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.txt)) that indicates that `time-offset` is optional. Of course, I may have overlooked something.

However, I found this quote (copied from https://json-schema.org/understanding-json-schema/reference/string#format):
> Implementations may provide validation for only a subset of the built-in formats or **do partial validation for a given format**.

It seems to offer the necessary wiggle room for this behavior to be considered "not a bug", i.e., while `"2023-01-01T12:34:56"` is not a valid `date-time` according to RFC 3339, implementations of json schema may still accept it as valid.

If the failing test is considered to be a bug I could look into creating a PR.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gregsdennis:
Also of note are:

- my [docs](https://docs.json-everything.net/schema/basics/#schema-format) on support for `format`:
[...]

PR #571, opened 2023-12-02
Make `time-offset` required for `date-time`

danielstarck:
Resolves https://github.com/gregsdennis/json-everything/issues/570.

The capturing group for `time-offset` (`([Zz]|[\+-]\d{2}:\d{2})`) has the quantifier `?` which allows it to occur zero times or once, effectively making it optional. Removing `?` ensures that it must occur exactly once.

### DIFF
--- a/JsonSchema.Tests/FormatTests.cs
+++ b/JsonSchema.Tests/FormatTests.cs
@@ -62,7 +62,7 @@ public class FormatTests
 
 		var result = schema.Evaluate(value, new EvaluationOptions { RequireFormatValidation = true });
 
-		Assert.False(result.IsValid);
+		result.AssertInvalid();
 	}
 
 	private static readonly Uri _formatAssertionMetaSchemaId = new("https://json-everything/test/format-assertion");
--- a/JsonSchema.Tests/FormatTests.cs
+++ b/JsonSchema.Tests/FormatTests.cs
@@ -52,6 +52,19 @@ public class FormatTests
 		Assert.True(result.IsValid);
 	}
 
+	[Test]
+	public void DateTime_MissingTimeOffset_Fail()
+	{
+		JsonSchema schema = new JsonSchemaBuilder()
+			.Format(Formats.DateTime);
+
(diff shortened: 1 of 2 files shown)


gregsdennis:
I'd be curious to see if the JSON Schema Test Suite had covers this. It'd be an optional test.

JsonSchema.Tests/FormatTests.cs
```diff
@@ -52,6 +52,19 @@ public void DateTime_Pass(string dateString)
...
+	public void DateTime_MissingTimeOffset_Fail()
+	{
+		JsonSchema schema = new JsonSchemaBuilder()
+			.Format(Formats.DateTime);
+
+		var value = JsonNode.Parse("\"2023-04-28T21:51:26\"");
+
+		var result = schema.Evaluate(value, new EvaluationOptions { RequireFormatValidation = true });
+
+		Assert.False(result.IsValid);
```
gregsdennis, line 65: Instead of a simple Assert here, please use the `.IsInvalid()` extension on the result.  This prints the schema and make it easier to debug if something goes wrong.

gregsdennis (changes requested):

danielstarck:
> I'd be curious to see if the JSON Schema Test Suite covers this. It'd be an optional test.

Can you elaborate? I'm not following. Are you asking that I take some action? :slightly_smiling_face:

gregsdennis:
These tests include the [official JSON Schema Test Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite).  I'm just curious if this is covered in one of the optional tests. They way I have it set up, the optional tests are inconclusive if they fail.

If it's not covered, then I think it would be a good addition.

You don't have to do anything if you don't want.  I can look later when the code's in front of me.

gregsdennis (approved):

gregsdennis:
Thanks for this @danielstarck!

gregsdennis:
I suppose I can't protect against regressions in the optional tests. I have to rely on reports for that.

Thank you.  Come join on the JSON Schema slack. We have a really good community growing.  The link is on json-schema.org.

PR #578, opened 2023-12-07
update for data-2023 and add path tests

gregsdennis:
Depends on https://github.com/gregsdennis/json-everything/pull/577
(doesn't really depend on it, but it has some other things that should ideally be merged first.)


gregsdennis:
Actually, I think I'm going to merge this into #577.  I need to release them both anyway, so I might as well figure out how to run them together now.

PR #586, opened 2023-12-12
Add type converters for JsonPointer and JsonPath

### LINKED ISSUE #585: Add TypeConverter for JsonPointer, JsonPath, etc. to allow easy configuration binding
### Nuget Package

JsonPointer.Net

### Package Version

3.2.2

When using the .NET configuration system, it is not currently possible to bind an entry from an `IConfiguration` to a `JsonPointer` on an options class. For example, the following always results in a null value for the `MyPointer` member on the `MyOptions` class:

```json
// appsettings.json
{
  "MyOptions": {
    "MyPointer": "/some/pointer"
  }
}
```

```csharp
// Application code

public class MyOptions {
  public JsonPointer MyPointer { get; set; }
}

// ...

void AddMyOptions(IServiceCollection services, IConfiguration configuration) {
  services.AddOptions<MyOptions>().Bind(configuration.GetSection("MyOptions"));
}
```

The .NET configuration binder uses [TypeConverter](https://github.com/dotnet/runtime/blob/e9ddf98c8502cfa00823dc94d3fcbfe5aa03ece7/src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs#L883)s to convert from strings to the target target type. By adding `TypeConverter` implementations for types such as `JsonPointer` and `JsonPath` it would allow options classes to easily make use of the types in this repository.

_No response_

_No response_

### Code of Conduct

wazzamatazz:
Closes #585

This PR adds `TypeConverter` implementations for `JsonPointer` and `JsonPath` to allow instances of these types to be created via Microsoft's configuration binder.

### DIFF
--- a/JsonPath.Tests/TypeConverterTests.cs
+++ b/JsonPath.Tests/TypeConverterTests.cs
@@ -1,0 +1,30 @@
+﻿using System.ComponentModel;
+using NUnit.Framework;
+
+namespace Json.Path.Tests
+{
+	public class TypeConverterTests
+	{
+
+		[Test]
+		public void ConvertFromString()
+		{
+			var typeConverter = TypeDescriptor.GetConverter(typeof(JsonPath));
+			var path = typeConverter.ConvertFromInvariantString("$.store.book[*].author") as JsonPath;
+
+			Assert.IsNotNull(path);
+			Assert.AreEqual("$.store.book[*].author", path!.ToString());
+		}
+
+		[Test]
+		public void ConvertToString()
+		{
+			var path = JsonPath.Parse("$.store.book[*].author");
+			var typeConverter = TypeDescriptor.GetConverter(typeof(JsonPath));
+			var pathString = typeConverter.ConvertToInvariantString(path);
+
+			Assert.AreEqual("$.store.book[*].author", pathString);
+		}
+
+	}
+}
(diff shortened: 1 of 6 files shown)


gregsdennis (changes requested):
Thanks for submitting this.

Please see https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.typeconverter?view=net-8.0#notes-to-inheritors

It looks like there are a few methods that should be implemented here since this is an immutable type.

It might also not be a bad idea to implement the standard value methods to support the empty case.  Not explicitly required, but it's probably a memory saver.

wazzamatazz:
> Thanks for submitting this.
> 
> Please see https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.typeconverter?view=net-8.0#notes-to-inheritors
> 
> It looks like there are a few methods that should be implemented here since this is an immutable type.
> 
> It might also not be a bad idea to implement the standard value methods to support the empty case. Not explicitly required, but it's probably a memory saver.

I've added implementations for various methods mentioned in the link and make some tweaks to the existing implementations. I referred to the [BCL code for UriTypeConverter](https://github.com/dotnet/runtime/blob/main/src/libraries/System.ComponentModel.TypeConverter/src/System/ComponentModel/UriTypeConverter.cs) when making the modifications, as conversion to/from `Uri` was the closest comparison that I could think of here.

Note that I've not implemented the `CreateInstance` or `GetCreateInstanceSupported` methods; `UriTypeConverter` does not implement these either.

gregsdennis:
Thanks for working this up.  I'm going to pull it into another branch to do all the release prep things.

Issue #390, opened 2023-03-01
Moving on from .Net Standard


gregsdennis:
From testing, it appears that updating to System.Text.Json v7 will require that I drop support for .Net Core 3.1.

I don't expect this will be a problem for most of my users as .Net Core 3.1 is [end-of-life](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core) anyway, but I feel it deserves notification and a major version bump in the libraries.

This will come... eventually.

m-gallesio:
Hello.
I just tried using this library in a .NET 7 project, and it pulled in System.Text.Json 7.0.0 with no apparent issues.
Unless there are specific reasons I don't think a forced upgrade is strictly required.

gregsdennis:
Some further discussions in #532 revealed that it might be time to move on from .Net Standard.  I need feedback from users.

This would probably apply to the entire suite.

gregsdennis:
I now have a branch with everything working for net6.0.

Might try to multi-target with net8.0 later, but very happy this is working for now.

gregsdennis:
This .Net 7 Regex stuff would be nice to have.

https://devblogs.microsoft.com/dotnet/regular-expression-improvements-in-dotnet-7/#stringsyntaxattribute-regex

PR #597, opened 2023-12-18
Remove Jetbrains dependency from the main library

RogerBarreto:
First I'm aware of the: Issue #174 

Althought this dependency don't necessarily needs to be in the main JsonSchema project, can be moved downstream to the projects that really depends on it.

image

As a good practice I suggest moving this to analyzers instead of annotations in the code. So this don't add a dependency that is not related to the actual purpose of the code-base.

### DIFF
--- a/JsonSchema/RegexFormat.cs
+++ b/JsonSchema/RegexFormat.cs
@@ -1,6 +1,5 @@
 ﻿using System.Text.Json.Nodes;
 using System.Text.RegularExpressions;
-using JetBrains.Annotations;
 
 namespace Json.Schema;
 
@@ -16,7 +15,7 @@ public class RegexFormat : Format
 	/// </summary>
 	/// <param name="key">The format key.</param>
 	/// <param name="regex">The regular expression.</param>
-	public RegexFormat(string key, [RegexPattern] string regex)
+	public RegexFormat(string key, string regex)
 		: base(key)
 	{
 		_regex = new Regex(regex, RegexOptions.ECMAScript | RegexOptions.Compiled);
--- a/JsonSchema.DataGeneration/JsonSchema.DataGeneration.csproj
+++ b/JsonSchema.DataGeneration/JsonSchema.DataGeneration.csproj
@@ -37,6 +37,7 @@
 	<ItemGroup>
 		<PackageReference Include="Bogus" Version="33.1.1" />
 		<PackageReference Include="Fare" Version="2.2.1" />
+		<PackageReference Include="JetBrains.Annotations" Version="2021.2.0" />
 	</ItemGroup>
 
 	<ItemGroup>
--- a/JsonSchema.Generation.Tests/JsonSchema.Generation.Tests.csproj
+++ b/JsonSchema.Generation.Tests/JsonSchema.Generation.Tests.csproj
@@ -14,6 +14,7 @@
 	</PropertyGroup>
 
 	<ItemGroup>
+		<PackageReference Include="JetBrains.Annotations" Version="2021.2.0" />
 		<PackageReference Include="NUnit" Version="3.13.3" />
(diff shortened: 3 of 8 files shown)


JsonSchema/RegexFormat.cs
```diff
@@ -16,7 +15,7 @@ public class RegexFormat : Format
 	/// </summary>
 	/// <param name="key">The format key.</param>
 	/// <param name="regex">The regular expression.</param>
-	public RegexFormat(string key, [RegexPattern] string regex)
+	public RegexFormat(string key, string regex)
```
gregsdennis, line 18: Removing this removes syntax highlighting for regex strings for consumers.  I find this very useful, and I expect other R# uses, too.

gregsdennis (changes requested):
These attributes have nothing to do with nullability or any of the other things listed in the screenshot.  Not everything provided by the package is covered by built-in analyzers.

I'm not sure why you're removing the JetBrains package from the main pkg but adding it to the extensions.

JsonSchema/RegexFormat.cs
RogerBarreto: This is an IDE XP thing, that may be only useful for devs using the R# in their environment, 

As of NET 7 moving forward we can use: `[StringSyntax(StringSyntaxAttribute.Regex)]`

As you can see in this blog post:

https://devblogs.microsoft.com/dotnet/regular-expression-improvements-in-dotnet-7

gregsdennis: Is that available as part of a .net standard 2.0 package? .net 7 features aren't available to these libraries.

RogerBarreto: > I'm not sure why you're removing the JetBrains package from the main pkg but adding it to the extensions.

To reduce the dependency graph, you will eventually only use this package if really need the `DataGeneration` library.

RogerBarreto: > Is that available as part of a .net standard 2.0 package? .net 7 features aren't available to these libraries.

Yes, its only for .Net 7+ libraries, don't apply here. The fewer dependencies the better + helps also increase adoption.

gregsdennis: I'm confused.  You're recommending that I replace the Jetbrains attribute with something that's only available in .net 7+, but these are .net standard 2.0 libs.

I'm happy to consider this as part of #390, but for now I'd prefer to stick with it as it is.

RogerBarreto: @gregsdennis, no, I did suggest it but didn't realized this attribute was not available to .Net Standard 2.0.

So, my suggestion would be to remove this dependency from the main package as a standard way of reducing the surface and dependencies that are not strictly necessary the project to work.

gregsdennis: It's a developer hint and convenience (dev UX is important to me).  I've had this conversation multiple times, and I'd like to have it.  If it's actually causing problems for you, I'm happy to talk about that, but I'm not going to remove it just at the mere suggestion.

In the end, they're non-functional, and the Jetbrains library is tiny (75 KB), so the impact to anything runtime is negligible.

RogerBarreto: I wont bring this further, the more dependencies you add to a package less likely you will see it being used across multiple places. Companies normally have compliance rules that restrict unecessary packages and this just adds more friction, but that's my view, my understanding. 

I'm dropping this now.

gregsdennis:
@RogerBarreto I appreciate the recommendation, but it just doesn't fit with what I want for the library at the time.  Thank you for your opinion and the pointer to the .Net 7 solution to the problem.  I'll definitely consider it when I make the move out of .Net Standard.

PR #598, opened 2023-12-18
sbyte is an integer

grbell-ms:


gregsdennis:
So it is!  I thought it odd that there wasn't a signed `byte` counterpart.

Ref: https://learn.microsoft.com/en-us/dotnet/api/system.sbyte

Seems that it's non-CLS-compliant, though, which has [implications](https://stackoverflow.com/a/1828616/878701) for libraries like this one.  I assume that the type isn't available in all .net languages.  I'm not sure that matters in this case, though, since I'd only be handling the cases where it is available. 🤔

gregsdennis:
@grbell-ms thanks for this!  I'll merge this into a publish branch to update versions and things.

PR #604, opened 2023-12-26
Performance improvement

danielstarck:
I was looking into improving the performance of my application so I set up a scenario with `BenchmarkDotNet` which, using that particular schema, evaluated 1000 json documents. These changes reduce the time from 465 ms to 220 ms.

Here are the results from running `JsonSchema.Benchmark` on my machine.

Before:
```
|          Method |  n |       Mean |   Error |  StdDev |       Gen0 |       Gen1 | Allocated |
|---------------- |--- |-----------:|--------:|--------:|-----------:|-----------:|----------:|
| RunSuite_Legacy |  1 |   523.1 ms | 5.43 ms | 4.82 ms | 10000.0000 |  5000.0000 |  127.6 MB |
| RunSuite_Legacy | 10 | 1,249.4 ms | 6.08 ms | 5.08 ms | 63000.0000 | 16000.0000 | 754.56 MB |
```

After:
```
|          Method |  n |     Mean |   Error |  StdDev |       Gen0 |      Gen1 | Allocated |
|---------------- |--- |---------:|--------:|--------:|-----------:|----------:|----------:|
| RunSuite_Legacy |  1 | 481.6 ms | 3.95 ms | 3.69 ms |  8000.0000 | 4000.0000 | 101.88 MB |
| RunSuite_Legacy | 10 | 793.2 ms | 5.69 ms | 5.33 ms | 42000.0000 | 5000.0000 | 506.87 MB |
```

### DIFF
--- a/JsonSchema/ErrorMessages.cs
+++ b/JsonSchema/ErrorMessages.cs
@@ -20,6 +20,7 @@ namespace Json.Schema;
 public static partial class ErrorMessages
 {
 	private static readonly ResourceManager _resourceManager = new("Json.Schema.Localization.Resources", typeof(ErrorMessages).Assembly);
+	private static readonly JsonSerializerOptions _jsonSerializerOptions = new JsonSerializerOptions { Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping };
 
 	/// <summary>
 	/// Gets or sets a culture to use for error messages.  Default is <see cref="CultureInfo.CurrentCulture"/>.
@@ -52,7 +53,7 @@ public static partial class ErrorMessages
 		for (var i = 0; i < parameters.Length; i++)
 		{
 			var parameter = parameters[i];
-			values[i] = JsonSerializer.Serialize(parameter.value, new JsonSerializerOptions { Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping });
+			values[i] = JsonSerializer.Serialize(parameter.value, _jsonSerializerOptions);
 			current = current.Replace($"[[{parameter.token}]]", $"{{{i}}}");
 		}
 
--- a/JsonSchema/Formats.cs
+++ b/JsonSchema/Formats.cs
@@ -26,6 +26,8 @@ public static class Formats
 		"HH':'mm':'ssK",
 		"HH':'mm':'ss"
 	};
+	
+	//from built from https://regex101.com/r/qH0sU7/1, edited to support all date+time examples in https://ijmacd.github.io/rfc3339-iso8601/
 	private static readonly Regex _dateTimeRegex = new Regex(@"^((?:(\d{4}-\d{2}-\d{2})([Tt_]| )(\d{2}:\d{2}:\d{2}(?:\.\d+)?))([Zz]|[\+-]\d{2}:\d{2}))$");
 
 	/// <summary>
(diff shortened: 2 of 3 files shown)


JsonSchema/ErrorMessages.cs
```diff
@@ -52,7 +53,7 @@ public static string ReplaceTokens(this string message, params (string token, ob
 		for (var i = 0; i < parameters.Length; i++)
 		{
 			var parameter = parameters[i];
-			values[i] = JsonSerializer.Serialize(parameter.value, new JsonSerializerOptions { Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping });
+			values[i] = JsonSerializer.Serialize(parameter.value, _jsonSerializerOptions);
```
danielstarck, line 56: STJ does some optimization when you re-use the same `JsonSerializerOptions`.

JsonSchema/Formats.cs
```diff
@@ -272,8 +273,7 @@ private static bool CheckDateFormat(JsonNode? node, params string[] formats)
 		//We use a fallback to catch these cases
 
 		//from built from https://regex101.com/r/qH0sU7/1, edited to support all date+time examples in https://ijmacd.github.io/rfc3339-iso8601/
-		var regex = new Regex(@"^((?:(\d{4}-\d{2}-\d{2})([Tt_]| )(\d{2}:\d{2}:\d{2}(?:\.\d+)?))([Zz]|[\+-]\d{2}:\d{2}))$");
-		var match = regex.Match(dateString);
+		var match = _dateTimeRegex.Match(dateString);
```
danielstarck, line 276: This is likely the reason there was a greater improvement for my scenario compared to `JsonSchema.Benchmark`, as I'm validating a date time in each call to `Evaluate`.

JsonSchema/JsonSchema.cs
```diff
@@ -405,7 +407,7 @@ private static SpecVersion DetermineSpecVersion(JsonSchema schema, SchemaRegistr
 	{
 		if (schema.BoolValue.HasValue) return SpecVersion.DraftNext;
 		if (schema.DeclaredVersion != SpecVersion.Unspecified) return schema.DeclaredVersion;
-		if (!Enum.IsDefined(typeof(SpecVersion), desiredDraft)) return desiredDraft;
```
danielstarck, line 408: I didn't quite understand the intention here, so I hope I got it right.

JsonSchema/Formats.cs
gregsdennis: Please move the comment on 275 up with the regex definition.

JsonSchema/JsonSchema.cs
gregsdennis: This actually supports an external library, my Graeae OpenAPI lib, by allowing it to define draft 4 as an enum value.  It's an undocumented hack, really, but it needs to work.

JsonSchema/JsonSchema.cs
```diff
@@ -510,6 +512,8 @@ private static void PopulateBaseUris(JsonSchema schema, JsonSchema resourceRoot,
 		}
 	}
 
+	private static bool IsDefinedSpecVersion(SpecVersion specVersion) => _specVersions.Contains(specVersion);
```
gregsdennis, line 515: I think I'd prefer not encapsulating this in a method since it's so simple. Just having the `.Contains()` call in the `if` statement reads fine, and it's not used in multiple places.
Maybe renaming `_specVersions` to `_definedSpecVersions` gets the message across.

gregsdennis (changes requested):

gregsdennis:
Thanks for looking into this.  These look like good optimizations overall.

gregsdennis:
I'm going to pull this into a release branch and do version / release note things.

gregsdennis (approved):

PR #608, opened 2023-12-29
Add cache for values derived from EvaluationOptions

danielstarck:
About 30% of the time taken for `Evaluate` is spent creating/mutating `EvaluationOptions`, `SchemaConstraint` and `EvaluationContext`. AFAICT this operation is dependent on the schema and the evaluation options, but not on the json document being evaluated. This means that consumers who always pass the same evaluation options (which I assume is quite common?) would benefit from having the result cached and re-used.

This draft is just supposed to illustrate the idea. I figured I'd get some feedback before spending more time on it. :slightly_smiling_face: 

`JsonSchema.Benchmark` before and after. I added a run with 100 iterations to allow the caching to shine.

Before:
```
|          Method |   n |       Mean |    Error |   StdDev |        Gen0 |       Gen1 |  Allocated |
|---------------- |---- |-----------:|---------:|---------:|------------:|-----------:|-----------:|
| RunSuite_Legacy |   1 |   462.1 ms |  3.37 ms |  2.99 ms |   8000.0000 |  4000.0000 |  102.46 MB |
| RunSuite_Legacy |  10 |   782.1 ms |  4.40 ms |  4.12 ms |  42000.0000 | 11000.0000 |  505.91 MB |
| RunSuite_Legacy | 100 | 3,726.5 ms | 18.29 ms | 17.10 ms | 378000.0000 |  5000.0000 | 4522.85 MB |
```

After:
```
|          Method |   n |       Mean |    Error |   StdDev |        Gen0 |      Gen1 |  Allocated |
|---------------- |---- |-----------:|---------:|---------:|------------:|----------:|-----------:|
| RunSuite_Legacy |   1 |   454.0 ms |  4.81 ms |  4.50 ms |   8000.0000 | 4000.0000 |   96.68 MB |
| RunSuite_Legacy |  10 |   672.1 ms |  5.86 ms |  5.20 ms |  31000.0000 | 2000.0000 |   377.2 MB |
| RunSuite_Legacy | 100 | 2,738.1 ms | 13.54 ms | 12.67 ms | 264000.0000 | 1000.0000 | 3164.95 MB |
```

I've also run some benchmarks with just `Evaluate` in a tight loop and they show around a 30% reduction in execution time and memory allocation. Let me know if you'd like me to share that.

(To be clear, I'm including the changes from https://github.com/gregsdennis/json-everything/pull/604 in my "Before" benchmarks, despite them not being in `master` just yet.)


JsonSchema/JsonSchema.cs
```diff
@@ -23,6 +24,7 @@ public class JsonSchema : IBaseDocument
 	
 	private readonly Dictionary<string, IJsonSchemaKeyword>? _keywords;
 	private readonly List<(DynamicScope Scope, SchemaConstraint Constraint)> _constraints = new();
+	private readonly ConcurrentDictionary<EvaluationOptions?, OptionsDerivedValues> _optionsDerivedValuesCache = new();
```
gregsdennis, line 27: Using the evaluation options as the key is basically a reference equality check (which may have hash collisions).  If a user is creating identical options objects, then the cache doesn't help anything.

danielstarck: >Using the evaluation options as the key is basically a reference equality check

Yes. 

>(which may have hash collisions)

I don't follow. Can you elaborate?

>If a user is creating identical options objects, then the cache doesn't help anything.

Right. This is how STJ `Serialize`/`Deserialize` works with regards to `JsonSerializerOptions`, so there is precedent for that kind of behavior. Whether you consider it good design is another question I suppose. 

More generally, is this kind of caching something you'd be in favor of? I'd be happy to look into adding support for the case where the consumer passes different objects that are equal by value.

gregsdennis: I think my concern is more around memory usage when the user doesn't reuse options. I'd be holding onto those objects unnecessarily.

danielstarck: I see. That's fair.

Maybe a dictionary is the wrong approach. My idea was to optimize for the case where the same options are re-used , and only a single cache entry is needed for that. I changed the code to only cache a single `OptionsDerivedValues` at a time. This satisfies my case while preventing the cache from growing over time. Does this seem like a better idea?

In order to get some idea of the size of the cache entry, I ran `GC.GetTotalMemory(true)` before and after creating and assigning it. The increase is 4952 bytes when using this schema. 

<details>
<summary>Schema</summary>

```
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://example.com/product.schema.json",
  "title": "Product",
  "description": "A product from Acme's catalog",
  "type": "object",
  "properties": {
    "productId": {
      "description": "The unique identifier for a product",
      "type": "integer"
    },
    "productName": {
      "description": "Name of the product",
      "type": "string"
    },
    "price": {
      "description": "The price of the product",
      "type": "number",
      "exclusiveMinimum": 0
    },
    "tags": {
      "description": "Tags for the product",
      "type": "array",
      "items": {
        "type": "string"
      },
      "minItems": 1,
      "uniqueItems": true
    },
    "dimensions": {
      "type": "object",
      "properties": {
        "length": {
          "type": "number"
        },
        "width": {
          "type": "number"
        },
        "height": {
          "type": "number"
        }
      },
      "required": [ "length", "width", "height" ]
    }
  },
  "required": [ "productId", "productName", "price" ]
}
```
</details>

danielstarck:
> * Identify options object by something rudimentary but still more than merely reference-equals so that people who continually recreate identical options can still benefit from this.  We may need to look into identifying which options actually affect evaluation itself (for example, `OutputFormat` doesn't affect evaluation).
Makes sense. I'll look into this.

> * Perhaps delegate the caching to a new class that implements an interface so that users can provide their own caching rules.
What kind of caching rules are you imagining?

gregsdennis:
What kind of caching rules are you imagining?

I don't know. I never know what my users will want. If I leave the actual cache itself to the user, it grants them more flexibility, and I can provide something basic that works.

danielstarck:
> * Identify options object by something rudimentary but still more than merely reference-equals so that people who continually recreate identical options can still benefit from this. 

Pushed some changes for this. Basically a value-equals.

> We may need to look into identifying which options actually affect evaluation itself (for example, `OutputFormat` doesn't affect evaluation).

I agree that `OutputFormat` has no impact on whether the document is considered valid. But it does affect the `EvaluationResults` that is returned, and AFAICT, that is the important part. Have I got that right?

For example: if a consumer changes their options from `OutputFormat.Flag` to `OutputFormat.List` between two calls to `Evaluate`, we'll have to get rid of the cached stuff and re-build it, or the returned `EvaluationResults` will still be created according to `OutputFormat.Flag`. 

> Perhaps delegate the caching to a new class that implements an interface so that users can provide their own caching rules.

I'm not sure what you're looking for here so I left this aside for now.

gregsdennis:
So output format isn't so concerning.  The things that I'm more concerned about are things that affect the constraints, like `.ValidateFormat` or `.EvaluateAs`.

Valdiating the `format` keyword occurs if either the `.ValidateFormat` option is set to true _or_ if the schema's meta-schema uses the `format-assertion` vocabulary.  Validating `format` or not creates different constraint sets.

`.EvaluateAs` allows the user to specify which JSON Schema version to process the schema under.  But this only has any effect if the schema itself doesn't specify that with a `$schema` keyword.  Different JSON Schema versions define different keywords, which affect the set of constraints that are built.

Another twist is that each options object can have its own schema and vocabulary registries.  This means that you can run the same schema aginst two different schema registries that have different schemas regsitered for the same `$id` value.  For example:

Schema:
```json
{
  "$id": "foo",
  "type": "object",
  "properties": {
    "bar": { "$ref": "bar" }
  }
}
```

Registry 1 contains:
```json
{
  "$id": "bar"
  "type": "integer"
}
```

Registry 2 contains
```json
{
  "$id": "bar"
  "type": "string"
}
```

These produce different constraint sets.

All of these are examples of things I don't really support well (if at all).  _That's_ what I want to fix.  Ironically, the previous architecture didn't have this problem.

gregsdennis:
I think the best approach would be to identify the things that could affect the constraint set and build new type that represents a state of those things to act as a cache key.  Looking at the options class, I see:

- `EvaluateAs`
- `SchemaRegistry`
- `VocabularyRegistry`
- `RequireFormatValidation`
- `OnlyKnownFormats` (throws for unknown formats)
- `ProcessCustomKeywords`

Accounting for the two registries will be challenging.  For the example above, the two schema registries have the same URI set.

The rest affect output, which can be recalculated each evaluation.

gregsdennis:
Another option is to just build constraints for everything in the schema, and then the evaluation step uses the options to filter out unneeded constraints.

I kinda like this option.

gregsdennis:
I tried making it generate constraints for everything today, but it wasn't working out.  I thought just creating a json pointer template to indicate unknown instance locations would do the trick, but it turned out much more complex than that.

gregsdennis:
@danielstarck I've retargeted this PR to a branch where I added some tests that these changes should pass.  Rightfully, the lib should pass these tests anyway, and currently they don't.  I need to fix this _somehow_.  If that's through this PR, cool; otherwise, I'll need to figure it out separately.

gregsdennis:
I've updated the branch to just clear the constraints when it detects a change in options that would affect validation outcome.  I don't think this PR is needed anymore.

gregsdennis:
I'm not sure what the cache is really doing in this PR.

- The constraints are really the only reusable information, and I'm already caching them, which is actually the cause of the problem for evaluation with different options.
- `EvaluationContext` doesn't hold anything that would be valuable between evaluations.  It's intentionally changing throughout the evaluation, and I don't guarantee it ends up in a reusable state.  I'm not really sure why it needs to be cached.  It's intentionally ephemeral.

danielstarck:
Sorry for disappearing. I haven't managed to find time for this. Glad to see you found a solution. :slightly_smiling_face:

Issue #460, opened 2023-05-24
Support `IParsable<T>`


gregsdennis:
**Environment**
- Nuget Package: JsonPath.Net, JsonPointer.Net
- .Net Target: .Net 7+

**Can this library make doing something simpler?**
Eventually, when this project [drops .NetStandard support](https://github.com/gregsdennis/json-everything/issues/310) and moves to the modern .Net versions, the [`IParsable<T>` interface](https://learn.microsoft.com/en-us/dotnet/api/system.iparsable-1?view=net-7.0) should be implemented on types that parse strings.

**Additional context**
This interface, and interface support for static methods, is only available for .Net 7 and higher.

gregsdennis:
This interface defines the pattern, but it also requires an `IFormatProvider` which has no meaning to these standards.

PR #611, opened 2024-01-15
Add [MemberNotNullWhen] to PatchResult

ivarne:
This removes the nullability warning when writing code like 

```C#
if(!result.IsSuccess)
{
   bool testOperationFailed = patchResult.Error.Contains("is not equal to the indicated value.");
}
```

This won't word before the library starts to (multi) target net5 or greater, but I still think it makes sense to have it ready as soon as that happens.

### DIFF
--- a/JsonPatch/PatchResult.cs
+++ b/JsonPatch/PatchResult.cs
@@ -23,6 +23,9 @@ public class PatchResult
 	/// <summary>
 	/// Gets whether there was an error.
 	/// </summary>
+	#if NET5_0_OR_GREATER
+	[System.Diagnostics.CodeAnalysis.MemberNotNullWhen(false, nameof(Error))]
+	#endif
 	public bool IsSuccess => Error == null;
 
 	internal PatchResult(PatchContext context)


ivarne:
> and that compiler directive will **never** be triggered.

OK, I just found https://github.com/gregsdennis/json-everything/issues/460 which seems to indicate that you were having plans to upgrade or multi target at some point.

gregsdennis:
Upgrade eventually, yes, but not multi-target.

gregsdennis:
I'm going to close this for now, but I'll definitely make these kinds of updates in the future.  Thanks for the suggestion.

gregsdennis:
@ivarne it looks like I might be moving toward updating sooner than I expected.  The demand for Native AOT support is growing.  If I need to do that I might as well do this as well.

See #618 for progress.

gregsdennis:
@ivarne I've retargeted this PR to `native-aot`, where this would be useful.  Yeah, I could just add the attribute myself, but I thought you'd like to have the commit.

ivarne:
Thanks!

This isn't my first PR, but if it was it would have mattered a lot.

PR #613, opened 2024-01-17
add tuple array conversion

gregsdennis:
Related to https://github.com/dotnet/runtime/issues/70352#issuecomment-1895750363

See a need, fill a need.

### DIFF
--- a/Json.More/JsonCollectionTupleConverter.cs
+++ b/Json.More/JsonCollectionTupleConverter.cs
@@ -1,0 +1,265 @@
+﻿using System;
+using System.Linq;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Json.More;
+
+internal class JsonCollectionTupleConverter : JsonConverterFactory
+{
+	public static JsonCollectionTupleConverter Instance { get; } = new();
+
+	public override bool CanConvert(Type typeToConvert)
+	{
+		return typeToConvert.FullName?.StartsWith("System.ValueTuple`") ?? false;
+	}
+
+	public override JsonConverter? CreateConverter(Type typeToConvert, JsonSerializerOptions options)
+	{
+		var typeParams = typeToConvert.GetGenericArguments();
+		Type converterType = typeParams.Length switch
+		{
+			1 => typeof(JsonCollectionTupleConverter<>).MakeGenericType(typeParams),
+			2 => typeof(JsonCollectionTupleConverter<,>).MakeGenericType(typeParams),
+			3 => typeof(JsonCollectionTupleConverter<,,>).MakeGenericType(typeParams),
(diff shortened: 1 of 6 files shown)


Json.More/JsonCollectionTupleConverter.cs
```diff
@@ -0,0 +1,265 @@
...
+
+		return (JsonConverter)Activator.CreateInstance(converterType, Array.Empty<object>());
+	}
+
+	public static JsonSerializerOptions ManageConverters(JsonSerializerOptions options)
+	{
+		var arrayConverters = options.Converters.OfType<JsonArrayTupleConverter>().ToArray();
+		if (arrayConverters.Length != 0)
+		{
+			options = new JsonSerializerOptions(options);
```
eiriktsarpalis, line 41: Curious what motivates inserting this logic -- as always creating a fresh JsonSerializerOptions forces recomputation of all serialization caches which is orders of magnitude more expensive than a regular serialization operation.

gregsdennis: I don't know of another way to swap out the converters (between the "array" and the "collection" one).  I also don't want to break other places the options objects I'm passed might be used.

gregsdennis: _A bit more detail (now that I'm on a PC and not my phone)._

You may notice that I actually have two converters.  This was done to solve the `TRest` overflow scenario.

`JsonArrayTupleConverter` reads the array bounds and uses `JsonCollectionTupleConverter` to read the values.  `JsonCollectionTupleConverter` can be called recursively.  Without this separation, the overflow is also serialized as an array rather than just tacking on additional values.

But since they both look for the same time, I can't just add `JsonCollectionTupleConverter` to the options' converter list; I have to replace `JsonArrayTupleConverter` with `JsonCollectionTupleConverter` to ensure the correct converter is used.

And I don't want to mutate the options object that the serializer gives me, so I make a copy only in the case where I need to perform the swap.

---

I did just realize, though, that having a deeply nested tuple probably won't work.

```c#
class MyObject
{
    public (string, int) Tuple { get; set; }
}

var obj = new MyObject { Tuple = ("foo", 42) };
var json = JsonSerializer.Serialize((false, obj), options);
```

I think this would likely not put the array bounds on the `Tuple` property...  I'll have to test this.

gregsdennis: Hmmm... yeah that's a problem.

```
Expected: "[false,{"Tuple":["foo",42]}]"
But was:  "[false,{"Tuple":"foo",42}]"
```

gregsdennis: I have an idea for a simpler solution.  I'll tag you in the PR.

PR #615, opened 2024-01-18
Better support for tuple serialization

gregsdennis:
There were some shortcomings with the original version of this converter set.  Apparently copying and modifying serializer options isn't ideal (I'll need to look at the validating converter also), and tuples with objects that had tuple properties would break.

cc: @eiriktsarpalis


Json.More/JsonArrayTupleConverter.cs
```diff
@@ -78,26 +75,18 @@ public override (T1, T2) Read(ref Utf8JsonReader reader, Type typeToConvert, Jso
...
-
-		options = JsonCollectionTupleConverter.ManageConverters(options);
-
-		var value = options.Read<(T1, T2)>(ref reader);
-		reader.Read();
 
-		if (reader.TokenType != JsonTokenType.EndArray)
-			throw new JsonException("Expected end of array.");
-		reader.Read();
+		var array = JsonElement.ParseValue(ref reader);
```
eiriktsarpalis, line 79: Why the need to allocate an intermediate JsonElement? Couldn't the value reader accept the `Utf8JsonReader` directly without making a copy?

Json.More/ValueWriter.cs
```diff
@@ -0,0 +1,105 @@
...
+		// ReSharper restore InconsistentNaming
+#pragma warning restore IDE1006
+	{
+		foreach (var item in values)
+		{
+			JsonSerializer.Serialize(writer, item.Item1, item.Item2, options);
+		}
+	}
+
+	public static IEnumerable<(object?, Type)> Unwrap1<T>(ValueTuple<T> value)
```
eiriktsarpalis, line 22: All tuple types implement the `ITuple` interface which provides an IList-like abstraction for accessing elements (and also handle `TRest` values correctly).

Json.More/JsonArrayTupleConverter.cs
```diff
@@ -48,26 +48,23 @@ public override ValueTuple<T> Read(ref Utf8JsonReader reader, Type typeToConvert
 	{
 		if (reader.TokenType != JsonTokenType.StartArray)
 			throw new JsonException("Expected tuple to be encoded as an array.");
-		reader.Read();
 
-		options = JsonCollectionTupleConverter.ManageConverters(options);
+		var array = JsonElement.ParseValue(ref reader);
```
eiriktsarpalis, line 52: Can't comment on the exact line, but it goes without saying that this JsonConverterFactory won't work in Native AOT since it's using unsupported reflection (Type.MakeGenericType). Unfortunately, a source generator is needed to be able to fold a nested tuple converter in a safe way.

Json.More/JsonArrayTupleConverter.cs
gregsdennis: I suppose so, yeah. I had initially needed to know the size of the array, but it seems I've iterated away from that.

Json.More/ValueWriter.cs
gregsdennis: `ITuple` doesn't exist in .Net Standard.  I looked at that.

Json.More/JsonArrayTupleConverter.cs
gregsdennis: Yeah, I'm still not sure how to get Native AOT working in .Net Standard 2.0.  There are other places I'd like to make things compatible.

Json.More/JsonArrayTupleConverter.cs
gregsdennis: Also, is it really an allocation if `JsonElement` and the enumerator are structs?

gregsdennis: So, it looks like because of the reflection, I can't use the reader directly because I need to pass it in by-ref, but I can't do that with `MethodInfo.Invoke()` because I can't put a by-ref into an object array.

If I can get around the reflection somehow, then I could do this, but I don't know how to get around it.

Once I hit more than 7 values, I have a `TRest` that I need to deal with.  I know that it's a `ValueTuple<...>`, but I don't know how many parameters there are, and that's where I need the reflection.  I can't do any static checks like `TRest is ValueType<...>` because I don't know what the `...` consists of or even how many type params there are.

Json.More/JsonArrayTupleConverter.cs
gregsdennis: I also have other areas that use reflection, like the `EnumStringConverter`.

Json.More/ValueWriter.cs
gregsdennis: Also, the ITuple interface returns `object`, but I wouldn't I need the type param to be sure I'm serializing the right type?  I could use the non-generic version of `JsonSerializer.Serialize()`, but that would just serialize what the _object_ is rather than what the _type parameter_ is, and they might be different.

gregsdennis:
@eiriktsarpalis I'm going to release this as-is since it's a good improvement over the current version.  If you know of a way to remove the reflection in .Net Standard 2.0, I'm happy to learn.

eiriktsarpalis:
FWIW this is not a .NET standard specific issue, there isn't a way to serialize arbitrarily nested tuples without using reflection that is not supported in Native AOT.

Issue #532, opened 2023-09-29
Parsing schema from string when compiled to NativeAOT with net8


jdtcn:
### Nuget Package

Json.More.Net

### Package Version

1.9.0

### Can this library make doing something simpler?

Hello!

Thanks for the cool library, that it is powered by System.Text.Json. Recently I tried to use this library in a net 8 application compiled in NativeAOT, and everything works if you do `JsonSchemaBuilder().FromType<T>().Build();`

Then I tried parsing the schema from the file as `JsonSchema.FromText(schema, options);` but this unfortunately doesn't work in NativeAOT, there are two problems:

* You need to create a custom converter for type `JsonConverter<Dictionary<string, JsonSchema>>` - that's pretty simple and after that the second problem occurs - 

* Method `Copy` In file `json-everything/Json.More/JsonNodeExtensions.cs:195` calls Deserialize without passing there `JsonSerializerOptions options` that are used to work in NativeAOT, and it fails `InvalidOperationException(JsonSerializerIsReflectionDisabled)`. This happens when calling `GetConstraint` from `ExamplesKeyword`.

It would be great if you could pass the JsonSerializerOptions options parameter in there, in this case your library will be able to parse schemas from string in NativeAOT mode.

Thanks!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gregsdennis:
Thanks for reporting this.  I'm not familiar with NativeAOT.  Would you be able to set up a reproduction with instructions to run it?

gregsdennis:
I've reimplemented `.Copy()` to not use the serializer.  Please try _Json.More.Net_ v1.9.1.

jdtcn:
Awesome, thanks for quick response! That helped, now Json.More.Net version 1.9.1 works with NativeAOT.

I've been experimenting further and found that there's another place that may be causing problems - `JsonSchema/ErrorMessages.cs:55`, there is a call to `JsonSerializer.Serialize` where a new instance of JsonSerializerOptions is created. But there is a workaround - add `<JsonSerializerIsReflectionEnabledByDefault>true</JsonSerializerIsReflectionEnabledByDefault>` in csproj, then everything will work.

But it would be cool to also inherit the JsonSerializerOptions in this place (NativeAOT needs TypeInfoResolver).

I created an example NativeAOT exe app, it will check the file against the schema, here it is - https://github.com/jdtcn/NativeAotJsonSchema

[...]

PR #618, opened 2024-01-24
Add multi-targeting for .NET 8

jevansaks:
To enable us to make opportunistic improvements to JsonSchema.NET for .NET 8 customers while keeping support for .NET Standard 2.0, this change adds multi-targeting support to all the projects.

I verified that consuming the new nupkg from a WPF or Net Standard 2.0 project gets the netstandard2.0 dll out of the nupkg, while a NET 8 project picks up the one from the net 8 folder.

Related to #390
Related to #532 (maybe resolves)

### DIFF
--- a/JsonSchema.Tests/JsonSchema.Tests.csproj
+++ b/JsonSchema.Tests/JsonSchema.Tests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFrameworks>netcoreapp3.1;net6.0</TargetFrameworks>
+		<TargetFrameworks>netcoreapp3.1;net8</TargetFrameworks>
 		<LangVersion>latest</LangVersion>
 
 		<IsPackable>false</IsPackable>
--- a/Json.More/Json.More.csproj
+++ b/Json.More/Json.More.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net8</TargetFrameworks>
     <PackageId>Json.More.Net</PackageId>
     <Authors>Greg Dennis</Authors>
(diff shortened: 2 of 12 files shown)


gregsdennis:
Hey @jevansaks thanks.  I'd be happy with you just adding onto this PR.  I'd prefer not just having this without the reason for doing it (supporting Native AOT).

I'm not afraid of a significant PR, and I understand this change will be considerable.

JsonSchema.Tests/JsonSchema.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFrameworks>netcoreapp3.1;net6.0</TargetFrameworks>
+		<TargetFrameworks>netcoreapp3.1;net8</TargetFrameworks>
```
gregsdennis, line 4: I'd actually prefer dropping 3.1 and having the current LTS versions here (6 & 8).

Also, can you update the other test projects, please?

jevansaks: Sure. So you'd like these projects that were `netcoreapp3.1;net6.0` to be `net6.0;net8`?

jevansaks: And anything that is netcoreapp3.1 should be net6;net8 too?

gregsdennis: Should only be tests that have .Net Core 3.1, but yes.

jevansaks: As long as we are continuing to support netstandard2.0 then we should continue to have netcoreapp3.1-based tests so that we can exercise both the reflection-based and JsonTypeInfo-based paths. We can have tests built and run 3x if you want to exercise netcoreapp3.1 _and_ net6 and net8.

gregsdennis:
I'm going to close this and replace with a PR from the `native-aot` branch.  It'll let me more easily work on this myself, and it includes some other things I've already done in working toward updating to .Net 6.

PR #619, opened 2024-01-24
Support Native AOT

### LINKED ISSUE #390: Moving on from .Net Standard

### LINKED ISSUE #532: Parsing schema from string when compiled to NativeAOT with net8

gregsdennis:
This PR adds support for Native AOT and source generation through implementing framework multitargeting.

Resolves #390
Resolves #532
Replaces #618 
Relates to #611

This is the root PR.  Since I have other people helping with this effort (thanks!), satellite PRs will be created to pull into this one.

To do:

- [x] Patch #627
- [x] Logic #629 
- [x] Pointer #624
- [x] ~[AOT test app](https://devblogs.microsoft.com/dotnet/creating-aot-compatible-libraries/) (section _Publishing a test application for AOT_)~ https://github.com/gregsdennis/json-everything/pull/626
- [x] ~Sanity unit tests to ensure all keywords are captured (used to be built with reflection)~
- [x] JsonSchema extension libs https://github.com/gregsdennis/json-everything/pull/635
  - [x] Data
  - [x] ArrayExt
  - [x] OpenApi
  - [x] DataGen
  - [x] SchemaGen (.Generation)
  - [x] CodeGen (I may drop this in favor of https://github.com/corvus-dotnet/Corvus.JsonSchema)
- [x] Version bumps (all major) & Release notes #638 
- [x] Docs
  - [ ] ~Add AOT-specific stuff~ (in separate repo; doesn't need to be done for this PR)
  - [x] ApiDocsGenerator to support net8
- [x] Code clean-up (personal preference)


gregsdennis:
☝️ this commit increases test runs to almost 52k!

jevansaks:
I was expecting to see the use of the JsonSerializerContext source generator and addition of `<IsAotCompatible>true</>` in the csproj files. What's the strategy here for ensuring that we're NAOT compatible?

gregsdennis:
> I was expecting to see the use of the JsonSerializerContext source generator and addition of `<IsAotCompatible>true</>` in the csproj files.

I'm getting there.

gregsdennis:
The majority of the tests that fail in [c9035e2](https://github.com/gregsdennis/json-everything/pull/619/commits/c9035e2cc8cc53714640db1752be2cd7980c49fa) are caused by https://github.com/dotnet/runtime/issues/97490.

Json.More/JsonArrayTupleConverter.cs
```diff
@@ -22,10 +23,12 @@ public override bool CanConvert(Type typeToConvert)
 	/// <param name="typeToConvert">The type handled by the converter.</param>
 	/// <param name="options">The serialization options to use.</param>
 	/// <returns>A converter which is compatible with <paramref name="typeToConvert" />.</returns>
+	[UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "<Pending>")]
+	[UnconditionalSuppressMessage("AOT", "IL3050:Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality when AOT compiling.", Justification = "<Pending>")]
```
jevansaks, line 27: These should  be changed to RequiresDynamicCode/RequiresUnreferencedCode so that we can let callers know this won't work.

gregsdennis: I tried that.  I get a warning that says the attributes have to match those on the base method.  Since the base method doesn't have those attributes, it doesn't like me adding them.  But it's okay with these.

No one's ever going to be calling the `.CreateConveter()` method direction anyway, so they'll never see the warnings if I put those attributes here.

JsonSchema/ErrorMessages.cs
```diff
@@ -19,44 +19,63 @@ namespace Json.Schema;
...
+    /// Replaces tokens in the form of `[[token]]` with a specified value, serialized as JSON.
+    /// </summary>
+    /// <param name="message">The message template.</param>
+    /// <param name="parameters">
+    /// Tuple of the token name (without brackets) and the value which will replace it.
+    /// </param>
+    /// <returns>The detokenized string.</returns>
+    [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "We guarantee that the SerializerOptions covers all the types we need for AOT scenarios.")]
+    [UnconditionalSuppressMessage("AOT", "IL3050:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "We guarantee that the SerializerOptions covers all the types we need for AOT scenarios.")]
+    public static string ReplaceTokens(this string message, params (string token, object? value)[] parameters)
```
gregsdennis, line 68: Technically we can't suppress these.  This is available to users, and they could pass in something we haven't covered.

We could change the params to (string, JsonNode?) perhaps.  That puts the serialization burden on the user, but that also means we'd have more call-site updates.  (Search for usages of `EvaluationResults.Fail()`.)

JsonSchema.Data/DataKeyword.cs
gregsdennis, line 1: So.... this is embarrassing.  This lib is supposed to also have an `optionalData` keyword that doesn't throw if references aren't resolved.  Pretty sure I coded it up, but I can't find it.

... and it's been released since the end of last year, supposedly with this keyword in it.

I'll add it.

JsonSchema.Data/UriIdentifier.cs
```diff
@@ -105,4 +110,20 @@ public override string ToString()
 	{
 		return Target.ToString();
 	}
+}
+
+[JsonSerializable(typeof(JsonSchema))]
+internal partial class DataGenerationSerializerContext : JsonSerializerContext
```
gregsdennis, line 116: @jevansaks why is this here?  copy/paste error?  There's one in the vocabularies file for this project already.

Json.More/JsonArrayTupleConverter.cs
jevansaks: Fair enough.

JsonSchema/ErrorMessages.cs
jevansaks: If we change this to RDC/RUC, does that balloon out to a bunch of call sites that need to be updated? It sounds like the right thing to do either way.

JsonSchema.Data/UriIdentifier.cs
jevansaks: Yeah, I admit I lost track sometimes of where I'd added JsonSerializerContexts. If there's another, let's switch to it.

gregsdennis: Taken care of in #638

JsonSchema/ErrorMessages.cs
gregsdennis: Yes, it would.  Could you do that in another PR, please?

jevansaks: Actually, to make that work we would need callers to pass in a IJsonTypeInfoResolver. I'm thinking we should add an AOT-friendly overload to handle it? Do you have an example of a caller using this with some types we didn't know about?

jevansaks: Now at https://github.com/gregsdennis/json-everything/pull/642

Json.More/AotCompatibleJsonConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+		/// <param name="value">The value to convert to JSON.</param>
+		/// <param name="options">An object that specifies serialization options to use.</param>
+		void Write(Utf8JsonWriter writer, object value, JsonSerializerOptions options);
+	}
+
+	/// <summary>
+	/// Abstract base class of JsonConverter<typeparamref name="T"/> that helps external code call 
+	/// a JsonConverter<typeparamref name="T"/> without statically knowing about T.
+	/// </summary>
+	public abstract class AotCompatibleJsonConverter<T> : JsonConverter<T>, IJsonConverterReadWrite
```
eiriktsarpalis, line 31: Nit: how about naming this `WeaklyTypedJsonConverter<T>`? It doesn't provide any AOT-specific guarantees.

Json.More/JsonArrayTupleConverter.cs
eiriktsarpalis: It should be possible to work around this issue by annotating with RUC/RDC at the type level, for example:

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Converters/Value/UnsupportedTypeConverterFactory.cs

Json.More/JsonElementExtensions.cs
```diff
@@ -134,6 +135,8 @@ void ComputeHashCode(JsonElement obj, ref int current, int depth)
 	/// <remarks>
 	/// See https://github.com/dotnet/runtime/issues/42502
 	/// </remarks>
+	[RequiresUnreferencedCode("Calls System.Text.Json.JsonSerializer.Serialize<T>(T)")]
```
eiriktsarpalis, line 138: Pretty sure the annotations could be avoided here -- these methods are dealing with known types (calls to `JsonSerializer` can be avoided altogether).

Json.More/JsonArrayTupleConverter.cs
gregsdennis: Nice. I didn't know you could annotation the type.  I'll add this.

Json.More/JsonArrayTupleConverter.cs
```diff
@@ -7,6 +8,9 @@ namespace Json.More;
 /// <summary>
 /// Provides JSON serialization for the <see cref="ValueTuple{T1}"/> family of types.
 /// </summary>
+/// <remarks>
```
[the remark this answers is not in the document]
gregsdennis: Nice. I didn't know you could annotation the type. I'll add this.

Json.More/JsonElementExtensions.cs
gregsdennis: Not sure how serializer calls could be avoided.  The point of this method is that `.ToString()` doesn't give you valid JSON, and `.GetRawText()` only gives the text that was input, including all of the whitespace; you can't perform other formatting.

If the serializer _can_ be avoided, I'm interested in that.  I mean, I could manually build it, but I prefer not to.

The other option here is to create a serializer context and toss `JsonElement` onto it, then just pass that in here.

Json.More/JsonElementProxy.cs
```diff
@@ -49,6 +50,8 @@ private JsonElementProxy(JsonElement value)
 	/// Converts an `string` to a <see cref="JsonElementProxy"/>.
 	/// </summary>
 	/// <param name="value">The value.</param>
+	[RequiresDynamicCode("Calls Json.More.JsonElementExtensions.AsJsonElement(string)")]
```
jevansaks, line 53: These attributes can be removed (they no longer call something that's annotated RDC/RUC).

JsonLogic/JsonLogic.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFramework>netstandard2.1</TargetFramework>
+		<TargetFrameworks>netstandard2.1;net8.0</TargetFrameworks>
```
jevansaks, line 4: I thought we were dropping netstandard with this change?

gregsdennis: No.  Actually, because of PolySharp, I might even be able to move back to NS 2.0!  I'll check that.

We dropped support for .Net Core 3.1 because of an incompatiblity in a dependency of STJ v8, but I still want to have .Net Standard for support outside of .Net 8.

gregsdennis: Looks like Logic can be pushed down to .Net Standard 2.0.  All I needed was to add a deconstructor for `KeyValuePair<,>`.

But Path is using all sorts of new APIs, like a lot of what's defined on spans.

jevansaks (approved):

PR #620, opened 2024-01-25
Make JsonSchema more AOT friendly

jevansaks:
The main change here is to use System.Text.Json's JsonSerializerContext source generators to produce JsonTypeInfos that we then use throughout the code for serialization. 

What library consumers need to know:
* JsonSchema methods that don't take a `JsonSerializerOptions` are AOT compatible by default
* JsonSchema methods that do take a `JsonSerializerOptions` can be AOT compatible if callers put `JsonSchema.TypeInfoResolver` in their `options.TypeInfoResolverChain`.
* Callers that dynamically register keywords must provide their own `JsonTypeInfo` via the new `SchemaKeywordRegistry.Register<T>(JsonTypeInfo)` overload, and that keyword's JsonConverter must implement `IJsonConverterReadWrite` or derive from `AotCompatibleJsonConverter<T>` instead of `JsonConverter<T>` (`AotCompatibleJsonConverter<T>` is just a convenience class that implements this interface for you).

Within the library:
* We continue to have JsonConverter implementations directly look up other JsonConverters via the JsonSerializerOptions to call their read methods directly. Instead of looking up via `options.GetConverter` (which is not AOT compatible), we use `options.GetTypeInfo` which is.
* JsonTypeInfo has the advantage that it has a Converter property directly on it. The problem for AOT is that the Read method is on `JsonConverter<T>` and we don't statically know all the `T` that there could be for the dynamically registered Keywords. The library today uses reflection to solve this, but for AOT we introduce an interface IJsonConverterReadWrite that we expect custom JsonConverters to implement. All our custom converters implement this (by deriving from `AotCompatibleJsonConverter<T>`) and we check that any dynamically registered Keyword through `Register<T>(JsonTypeInfo)` implements the interface as well.
* I had to do some `#if NET8_OR_GREATER` checks around certain things that are only available in System.Text.Json 8.x, like `JsonTypeInfo.Converter`. Thankfully these are relatively few.
* There were some instances of self-reflection to enumerate the list of keywords. I manually listed them in one place and the other places reference the one list from SchemaKeywordRegistry.


JsonSchema.Tests/VocabularyTests.cs
```diff
@@ -136,7 +136,7 @@ public MaxDateKeyword(DateTime date)
 		}
 	}
 
-	public class MaxDateJsonConverter : JsonConverter<MaxDateKeyword>
+	private class MaxDateJsonConverter : JsonConverter<MaxDateKeyword>
```
gregsdennis, line 139: I don't mind everything in the tests being public.

JsonSchema/Formats.cs
```diff
@@ -113,11 +114,27 @@ public static class Formats
 
 	static Formats()
 	{
-		_registry = new ConcurrentDictionary<string, Format>(
-			typeof(Formats)
-				.GetFields(BindingFlags.Static | BindingFlags.Public)
-				.Select(f => (Format)f.GetValue(null)!)
-				.ToDictionary(f => f.Key));
+		_registry = new ConcurrentDictionary<string, Format>(new Dictionary<string, Format>() {
```
gregsdennis, line 117: Can we do without instantiating the internal dictionary (just use the concurrent one)?

JsonSchema/JsonSchema.cs
```diff
@@ -21,7 +106,11 @@ public class JsonSchema : IBaseDocument
 {
 	private const string _unknownKeywordsAnnotationKey = "$unknownKeywords";
 
+#if NET6_0_OR_GREATER
+	private static readonly HashSet<SpecVersion> _definedSpecVersions = [..Enum.GetValues<SpecVersion>()];
+#else
 	private static readonly HashSet<SpecVersion> _definedSpecVersions = [..Enum.GetValues(typeof(SpecVersion)).Cast<SpecVersion>()];
+#endif
```
gregsdennis, line 113: Can we move the enum value fetching into a helper function somewhere?  It would clean this up a bit.  We get enum values elsewhere also.

JsonSchema/SchemaKeywordRegistry.cs
```diff
@@ -18,40 +20,126 @@ namespace Json.Schema;
 public static class SchemaKeywordRegistry
 {
 	private static readonly ConcurrentDictionary<string, Type> _keywords;
+	private static readonly ConcurrentDictionary<Type, JsonTypeInfo> _keywordTypeInfos;
 	private static readonly ConcurrentDictionary<Type, IJsonSchemaKeyword> _nullKeywords;
+	internal static bool RequiresDynamicSerialization { get; private set; }
```
gregsdennis, line 25: I don't think this needs to be set globally.  Can we make it per keyword and have a method that gets whether the particular keyword needs to be dynamic?

JsonPointer/JsonPointer.csproj
```diff
@@ -33,6 +33,7 @@
 	<Nullable>enable</Nullable>
 	<NoWarn>IDE0290</NoWarn>
 	<PolySharpIncludeRuntimeSupportedAttributes>true</PolySharpIncludeRuntimeSupportedAttributes>
+    <IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">true</IsAotCompatible>
```
gregsdennis, line 36: If we put a public serializer context here, wouldn't that mean we can use that in other places, e.g. JsonSchema?

JsonSchema/GlobalSuppressions.cs
gregsdennis, line 1: Will these be cleaned up?  I expect not all of them are still necessary with all of the changes.

JsonSchema/JsonSchema.cs
```diff
@@ -1,17 +1,102 @@
...
+[JsonSerializable(typeof(Dictionary<string, JsonSchema>))]
+[JsonSerializable(typeof(Dictionary<string, bool>))]
+[JsonSerializable(typeof(List<JsonSchema>))]
+[JsonSerializable(typeof(List<string>))]
+[JsonSerializable(typeof(JsonArray))]
+[JsonSerializable(typeof(Dictionary<string, SchemaOrPropertyList>))]
+[JsonSerializable(typeof(Dictionary<string, List<string>>))]
+[JsonSerializable(typeof(int[]))]
+[JsonSerializable(typeof(Dictionary<string, PropertyDependency>))]
+internal partial class JsonSchemaSerializationContext : JsonSerializerContext
```
gregsdennis, line 95: We had talked about making this public and only carry the public types with another for the internal types.

JsonSchema/JsonSchema.cs
```diff
@@ -709,8 +852,27 @@ public override JsonSchema Read(ref Utf8JsonReader reader, Type typeToConvert, J
...
+						{
+							implementation = JsonSerializer.Deserialize(ref reader, typeinfo!) as IJsonSchemaKeyword ??
+								throw new InvalidOperationException($"Could not deserialize expected keyword `{keyword}`");
+						}
+						else
+#endif
+						if (SchemaKeywordRegistry.RequiresDynamicSerialization)
+						{
+#pragma warning disable IL2026, IL3050 // Suppress because to get here the caller of SchemaKeywordRegistry must have suppressed the AOT warnings themselves
+							implementation = JsonSerializer.Deserialize(ref reader, keywordType) as IJsonSchemaKeyword ??
```
gregsdennis, line 867: Wouldn't we still want the `options.Read()` here?

JsonSchema/SchemaKeywordRegistry.cs
```diff
@@ -94,6 +182,20 @@ public static void RegisterNullValue<T>(T nullKeyword)
...
+
+	/// <summary>
+	/// Get the JsonTypeInfo for a given keywordType
+	/// </summary>
+	/// <param name="keywordType">Type to look up</param>
+	/// <param name="jsonTypeInfo">JsonTypeInfo for serialization</param>
+	/// <remarks>
+	/// This supports callers that use the JsonTypeInfo overloads of JsonSerializer methods.
+	/// </remarks>
+	public static bool TryGetTypeInfo(Type keywordType, out JsonTypeInfo? jsonTypeInfo)
```
gregsdennis, line 194: We had discussed making this internal.

JsonLogic/Rule.cs
```diff
@@ -72,7 +72,7 @@ public abstract class Rule
 /// <summary>
 /// Provides serialization for all <see cref="Rule"/> derivatives.
 /// </summary>
-public class LogicComponentConverter : JsonConverter<Rule>
+public class LogicComponentConverter : JsonConverter<Rule>, Json.More.IJsonConverterReadWrite<Rule>
```
gregsdennis, line 75: Why are we adding this to all of the converters?  I thought this was just to support the `object? Read()` used in JsonSchema.

This interface never gets used on these converters.

JsonPatch/JsonPatch.cs
```diff
@@ -116,10 +115,6 @@ public override JsonPatch Read(ref Utf8JsonReader reader, Type typeToConvert, Js
 	/// <param name="options">An object that specifies serialization options to use.</param>
 	public override void Write(Utf8JsonWriter writer, JsonPatch value, JsonSerializerOptions options)
 	{
-#if NET6_0_OR_GREATER
-		JsonSerializer.Serialize(writer, value.Operations, JsonPatchSerializerContext.Default.IReadOnlyListPatchOperation);
-#else
-		JsonSerializer.Serialize(writer, value.Operations);
-#endif
+		JsonSerializer.Serialize(writer, value.Operations, options);
```
gregsdennis, line 118: I don't understand.  Why are we not including the JTI now?

JsonPointer/JsonPointer.csproj
jevansaks: Is this comment in the right place? I'm not sure which context you mean.

JsonSchema.Tests/VocabularyTests.cs
jevansaks: I put this back.

JsonLogic/Rule.cs
jevansaks: I'll revert these changes.

JsonSchema/JsonSchema.cs
jevansaks: Yeah, I've gone back to the options.Read pattern which loses the static analysis but makes sure that this behaves the same. With the amount of test coverage that's here, I think we'll be ok.

JsonSchema/Formats.cs
jevansaks: ConcurrentDictionary doesn't have a constructor that is compatible with object initialization (it implements the Add method explicitly so the compiler doesn't find it). Do you have a suggestion for how to rewrite it?

JsonSchema/SchemaKeywordRegistry.cs
jevansaks: I can scope it to just the dynamic keywords.

JsonPatch/JsonPatch.cs
jevansaks: I'm putting the Serialize methods back the way they were. JsonSerializer.Serialize will find the JTIs through the options class.

JsonSchema/JsonSchema.cs
jevansaks: Can do.

JsonSchema/Formats.cs
gregsdennis: No, I guess this is fine.  A bit annoying but fine.

JsonSchema/SchemaKeywordRegistry.cs
gregsdennis: No, that wouldn't catch custom keywords.  If I get complaints, I'll change it.

JsonPointer/JsonPointer.csproj
gregsdennis: We discussed in Slack.  Nevermind.

JsonSchema/JsonSchemaSerializerContext.cs
```diff
@@ -0,0 +1,90 @@
...
+	[JsonSerializable(typeof(Dictionary<string, JsonSchema>))]
+	[JsonSerializable(typeof(Dictionary<string, bool>))]
+	[JsonSerializable(typeof(List<JsonSchema>))]
+	[JsonSerializable(typeof(List<string>))]
+	[JsonSerializable(typeof(JsonArray))]
+	[JsonSerializable(typeof(Dictionary<string, SchemaOrPropertyList>))]
+	[JsonSerializable(typeof(Dictionary<string, List<string>>))]
+	[JsonSerializable(typeof(int[]))]
+	[JsonSerializable(typeof(Dictionary<string, PropertyDependency>))]
+	internal partial class JsonSchemaSerializerContext : JsonSerializerContext
```
gregsdennis, line 86: We had talked about splitting this into two contexts, one for public types and one for internal.

gregsdennis: @eiriktsarpalis, what's the guidance for exposing serializer contexts?

JsonSchema/GlobalSuppressions.cs
jevansaks: I added a comment:

```
// Unfortunately the overload of JsonSerializer.Serialize that these are calling has the AOT warning attributes,
// so these suppressions are required. These calls are AOT compatible as long as the options object contains a
// TypeInfoResolver that covers all the types we need, which should be the case for the AOT paths. I've suppressed
// the warnings for now. A better solution might be to switch the JsonConverters to fetch the nested types'
// JsonTypeInfo.Converter and call Write directly, just like the Read methods do. I left it for now to reduce churn.
```

JsonSchema/JsonSchemaSerializerContext.cs
jevansaks: What Eirik told me is that only the TypeInfoResolver is needed so that callers can add it to their JsonSerializerOptions.TypeInfoResolverChain. I don't know if there's a pattern for this yet, but I've exposed JsonSchema.TypeInfoResolver public property to cover this.

gregsdennis:
This looks great.  Thanks for putting it together.  Still quite a bit to do around the JsonSchema extension libs, and of course docs, release notes, etc.

PR #622, opened 2024-01-25
update jsonpath for native aot

gregsdennis:
Updates JsonPath.Net for NativeAOT support

### DIFF
--- a/JsonPath/IPathFunctionDefinition.cs
+++ b/JsonPath/IPathFunctionDefinition.cs
@@ -27,9 +27,15 @@ internal interface IReflectiveFunctionDefinition
 /// <summary>
 /// Base class for defining an expression function which returns `ValueType`.
 /// </summary>
-public abstract class ValueFunctionDefinition : IReflectiveFunctionDefinition, IPathFunctionDefinition
+public abstract partial class ValueFunctionDefinition : IReflectiveFunctionDefinition, IPathFunctionDefinition
 {
-	internal class NothingValue;
+	private class NothingValue;
+
+	[JsonSerializable(typeof(NothingValue))]
+	private partial class NothingValueContext : JsonSerializerContext
+	{
+
+	}
 
 	/// <summary>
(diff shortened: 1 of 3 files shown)


JsonPath/IPathFunctionDefinition.cs
```diff
@@ -33,6 +34,10 @@ private class NothingValue{}
 	/// <summary>
 	/// Represents the absence of a JSON value and is distinct from any JSON value, including null.
 	/// </summary>
+	[UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code",
+		Justification = "This type is not returned to the user and should never be serialized.")]
+	[UnconditionalSuppressMessage("AOT", "IL3050:Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality when AOT compiling.",
+		Justification = "This type is not returned to the user and should never be serialized.")]
 	public static JsonValue Nothing { get; } = JsonValue.Create(new NothingValue())!;
```
jevansaks, line 41: There's an AOT-friendly way to do this, if you get a JsonTypeInfo of NothingValue then you can go through an overload of JsonValue.Create that doesn't trip the warnings.

Something like this would do it:
```C#
/// <summary>
/// Base class for defining an expression function which returns `ValueType`.
/// </summary>
public abstract partial class ValueFunctionDefinition : IReflectiveFunctionDefinition, IPathFunctionDefinition
{
	private class NothingValue{}

	[JsonSerializable(typeof(NothingValue))]
	private partial class NothingValueSerializerContext : JsonSerializerContext
	{

	}

	/// <summary>
	/// Represents the absence of a JSON value and is distinct from any JSON value, including null.
	/// </summary>
	public static JsonValue Nothing { get; } = JsonValue.Create(new NothingValue(), NothingValueSerializerContext.Default.NothingValue)!;

```

jevansaks (changes requested):

jevansaks (approved):

PR #624, opened 2024-01-27
fix up json pointer (and a couple others)

gregsdennis:


JsonSchema/JsonSchemaSerializerContext.cs
```diff
@@ -71,6 +72,8 @@ namespace Json.Schema
 	[JsonSerializable(typeof(UnrecognizedKeyword))]
 	[JsonSerializable(typeof(VocabularyKeyword))]
 	[JsonSerializable(typeof(WriteOnlyKeyword))]
+	[JsonSerializable(typeof(EvaluationResults))]
+	[JsonSerializable(typeof(JsonPointer))]
```
gregsdennis, line 76: Would like to find a way to just add in the type resolver exposed by JsonPointer rather than have to regen it.

@eiriktsarpalis, any ideas?

jevansaks (approved):

JsonSchema/JsonSchemaSerializerContext.cs
eiriktsarpalis: Assuming you're only looking to serialize `JsonPointer` as a root-level value (and not types _containing_ `JsonPointer` as a nested values which necessitates regeneration), you could try using [this combinator](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.metadata.jsontypeinforesolver.combine?view=net-8.0). It can be combined like so:
```C#
JsonTypeInfoResolver.Combine(ContextA.Default, ContextB,Default);
```

gregsdennis: I was looking for a way to incorporate it directly into the serializer context.

We eventually found a way using the serializer options.  We've extracted it into what is currently called `TypeResolverOptionsManager` in Json.More.  It's not the most elegant right now, but it works by creating a serializer options that includes dependency type resolvers.

PR #625, opened 2024-01-27
Fix ErrorMessages serialization for AOT

jevansaks:
When testing out real AOT, I uncovered one suppression that was covering a real issue. This call to Serialize was passing a custom JsonSerializerOptions that didn't include the TypeInfoResolver. Adding that revealed that keywords weren't serializing without a couple of extra primitive types that they depended on (`decimal`, `uint`, `JsonNode[]`, etc).

This revealed that we weren't including the custom TypeInfos for keywords during that serialization either, so to fix that we need to construct a JsonSerializerOptions that has the chain from our own as well as any callers.

I've changed the pattern so that we call serialize/deserialize with a custom Options now that's based on our source generated context but can dynamically get the context from any dynamically registered keywords too.


JsonSchema/SchemaKeywordRegistry.cs
```diff
@@ -21,78 +22,82 @@ namespace Json.Schema;
 public static class SchemaKeywordRegistry
 {
 	private static readonly ConcurrentDictionary<string, Type> _keywords;
-	private static readonly ConcurrentDictionary<Type, JsonTypeInfo> _keywordTypeInfos;
+	private static readonly ConcurrentDictionary<Type, JsonSerializerContext> _keywordTypeInfoResolvers = new();
```
gregsdennis, line 25: It doesn't look like this is ever initialized with the default keyword set.

JsonSchema/JsonSchema.cs
```diff
@@ -91,12 +92,96 @@ public class JsonSchema : IBaseDocument
...
+					TypeInfoResolver = _typeInfoResolver
+#endif
+				};
+
+				return _serializerOptions!;
+			}
+		}
+	}
+
+	internal static JsonSerializerOptions SerializerOptionsUnsafeRelaxedJsonEscaping
```
gregsdennis, line 114: It doesn't look like these things actually have to do with the `JsonSchema` class specifically.  I'd like them removed into a separate (static) class, please.

JsonSchema/SchemaKeywordRegistry.cs
jevansaks: It isn't, those are from the default JsonSchema serializer context. I can add a comment.

JsonSchema/JsonSchema.cs
jevansaks: Sure.

JsonSchema/SchemaKeywordRegistry.cs
gregsdennis: Okay... this is importing new contexts.  Got it.

JsonSchema/JsonSchema.cs
jevansaks: Would you be ok if they're on the JsonSchemaSerializerContext?

gregsdennis: Yeah, if that won't interfere with anything else.

gregsdennis:
Looks great!

PR #626, opened 2024-01-27
Native aot test app

gregsdennis:
Depends on #624

### DIFF
--- a/AotCompatibility/AotCompatibility.csproj
+++ b/AotCompatibility/AotCompatibility.csproj
@@ -1,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+	<InvariantGlobalization>true</InvariantGlobalization>
+	  
+    <PublishAot>true</PublishAot>
+	<TrimmerSingleWarn>false</TrimmerSingleWarn>
+	<SelfContained>true</SelfContained>
+  </PropertyGroup>
+
+  <ItemGroup>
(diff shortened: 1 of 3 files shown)


AotCompatibility/AotCompatibility.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+	<InvariantGlobalization>true</InvariantGlobalization>
+	  
+    <PublishAot>true</PublishAot>
```
jevansaks, line 10: Add:
```
		<IsTrimmable>true</IsTrimmable>
```

gregsdennis: Doesn't `PublishAot` cover that?

gregsdennis: https://learn.microsoft.com/en-nz/dotnet/core/deploying/trimming/prepare-libraries-for-trimming?pivots=dotnet-8-0#enable-project-specific-trimming

> The IsTrimmable property defaults to true when configuring a project as AOT-compatible with `<IsAotCompatible>true</IsAotCompatible>`.

Do we need to set that separately from `PublishAot`?

jevansaks: I'm not sure. @sergio0694 always tells me to include it, perhaps he can give better guidance.

Sergio0694: If you use `PublishAot`, that's sufficient. `IsAotCompatible` is used for libraries, but if you're making an app, the former covers everything you need already 🙂

gregsdennis:
Closing this for now.  I think forbidding reflection in the tests is sufficient.

PR #627, opened 2024-01-27
Patch updates

gregsdennis:
Was going to do both patch and logic here, but decided not to.

### DIFF
--- a/JsonPatch/JsonPatch.cs
+++ b/JsonPatch/JsonPatch.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Text.Json;
 using System.Text.Json.Nodes;
 using System.Text.Json.Serialization;
+using System.Text.Json.Serialization.Metadata;
 using Json.More;
 using Json.Pointer;
 
@@ -15,6 +16,14 @@ namespace Json.Patch;
 [JsonConverter(typeof(PatchJsonConverter))]
 public class JsonPatch : IEquatable<JsonPatch>
 {
+#if NET8_0_OR_GREATER
+	/// <summary>
+	/// A TypeInfoResolver that can be used for serializing <see cref="JsonPointer"/> objects. Add to your custom
+	/// JsonSerializerOptions's TypeInfoResolver or TypeInfoResolveChain.
+	/// </summary>
+	public static IJsonTypeInfoResolver JsonTypeResolver => PatchSerializerContext.Default;
+#endif
+
 	/// <summary>
 	/// Gets the collection of operations.
 	/// </summary>
@@ -86,7 +95,7 @@ public class JsonPatch : IEquatable<JsonPatch>
 /// <summary>
 /// Provides JSON conversion logic for <see cref="JsonPatch"/>.
 /// </summary>
-public class PatchJsonConverter : JsonConverter<JsonPatch>
+public class PatchJsonConverter : AotCompatibleJsonConverter<JsonPatch>
 {
 	/// <summary>Reads and converts the JSON to type <see cref="JsonPatch"/>.</summary>
 	/// <param name="reader">The reader.</param>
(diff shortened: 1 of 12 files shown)


JsonPatch/PatchOperation.cs
```diff
@@ -159,7 +159,7 @@ private class Model
 
 	public override PatchOperation Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
 	{
-		var model = JsonSerializer.Deserialize<Model>(ref reader, options)!;
+		var model = JsonSerializer.Deserialize(ref reader, PatchSerializerContext.Default.Model)!;
```
gregsdennis, line 162: Not really sure why, but this isn't deserializing the contents.  I get an empty model instance.

jevansaks: Does Model have a JsonConverter? If not, take a look at the generated code for PatchSerializerContext and see what it produces. I looked at this and the serializer it generated for Model was just empty. Maybe there are some attributes you can add to Model that help the JsonSerializable code generator know how to serialize the fields?

gregsdennis: I would expect the serializer to be able to handle POCOs out of the box.  Strange.  I'll try adding a custom converter and see if it makes a difference.

jevansaks: Is there a behavior difference in net8 versus netcoreapp3.1 for the tests? Could be a bug in STJ 6, I've seen a bunch of issues like that. I'll build your branch locally and look too.

jevansaks: Also I agree, the documentation says that it should be able to generate the serializers for primitive types.

jevansaks: This is weird. I'm debugging "PatchExtensionTests.Add" test and there's a bunch of Model deserializes that work fine and then the one that the test passes in doesn't deserialize. I haven't figured it out yet.

jevansaks: Looks like the Model's properties are "Op" and the test is providing a payload where the properties are "op". I suspect Options.PropertyNameCaseInsensitive is coming into play here but I haven't finished tracking it down.

gregsdennis: I bet you're right: case sensitivity.  I can mark with `[JsonProperty()]`.  That will likely fix it.  Good thinking.

JsonPatch/JsonPatch.cs
```diff
@@ -15,6 +16,14 @@ namespace Json.Patch;
 [JsonConverter(typeof(PatchJsonConverter))]
 public class JsonPatch : IEquatable<JsonPatch>
 {
+#if NET8_0_OR_GREATER
+	/// <summary>
+	/// A TypeInfoResolver that can be used for serializing <see cref="JsonPointer"/> objects. Add to your custom
+	/// JsonSerializerOptions's TypeInfoResolver or TypeInfoResolveChain.
+	/// </summary>
+	public static IJsonTypeInfoResolver JsonTypeResolver => PatchSerializerContext.Default;
```
jevansaks, line 24: I called this TypeInfoResolver in JsonSchema. I don't have a strong preference, but we should be consistent.

JsonPatch/JsonPatch.cs
```diff
@@ -95,7 +104,7 @@ public class PatchJsonConverter : JsonConverter<JsonPatch>
 	/// <returns>The converted value.</returns>
 	public override JsonPatch Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
 	{
-		var operations = JsonSerializer.Deserialize<List<PatchOperation>>(ref reader, new JsonSerializerOptions { PropertyNamingPolicy = JsonNamingPolicy.CamelCase })!;
```
jevansaks, line 98: This is where we lost the CamelCase (and presumably the case insensitivity) from before. You can add it back through the JsonSourceGeneratorOptions attribute on the PatchSerializerContext if that makes sense? But I also saw you did name aliases on the Model objects, that might be cleaner.

jevansaks (approved):
Looks good, left a couple suggestions.

PR #628, opened 2024-01-28
Native aot type safe read

gregsdennis:
Updated the `options.Read<T>()` extension to take a `JsonTypeInfo<T>` instead of the non-generic one.

This change infers the type parameter when a JTI is passed, which improves readability while also allowing us to explicitly use our serializer context.  Additionally, it adds a compiler check to ensure that we're generating a JTI for every type that we explicitly deserialize.


jevansaks:
Cool! I like it.

jevansaks (approved):

PR #629, opened 2024-01-28
Native aot logic

gregsdennis:
Depends on #628


gregsdennis:
Getting a few oddities

- can't deserialize `null` into a rule
- building values using the nodes API doesn't work

```c#
var array = new JsonArray { 1, 2, 3 };
```

apparently throws a serialization exception that int hasn't been registered on the serialization context!

stack trace:

```
System.NotSupportedException : JsonTypeInfo metadata for type 'System.Int32' was not provided by TypeInfoResolver of type 'System.Text.Json.Serialization.Metadata.EmptyJsonTypeInfoResolver'. If using source generation, ensure that all root types passed to the serializer have been annotated with 'JsonSerializableAttribute', along with any types that might be serialized polymorphically.
   at System.Text.Json.ThrowHelper.ThrowNotSupportedException_NoMetadataForType(Type type, IJsonTypeInfoResolver resolver)
   at System.Text.Json.JsonSerializerOptions.GetTypeInfoInternal(Type type, Boolean ensureConfigured, Nullable`1 ensureNotNull, Boolean resolveIfMutable, Boolean fallBackToNearestAncestorType)
   at System.Text.Json.JsonSerializerOptions.GetTypeInfo(Type type)
   at System.Text.Json.Nodes.JsonNode.ConvertFromValue[T](T value, Nullable`1 options)
   at System.Text.Json.Nodes.JsonArray.Add[T](T value)
```

@eiriktsarpalis, can you shed any light on this?

JsonLogic/Rule.cs
```diff
@@ -136,7 +146,7 @@ public override void Write(Utf8JsonWriter writer, Rule value, JsonSerializerOpti
 	{
 		if (value.Source != null)
 		{
-			JsonSerializer.Serialize(writer, value.Source, options);
+			JsonSerializer.Serialize(writer, value.Source, LogicSerializerContext.Default.JsonNode);
```
jevansaks, line 149: I don't know the context that these JsonConverters are being used, but I think we want a pattern in the Write methods that's similar to the read -- that is, respecting the full TypeInfoResolver chain that's in the JsonSerializerOptions passed in.

By switching from "options" to the JsonSerializerContext from our module only, we prevent a caller from extending the list.

As we discussed in Slack, I think we should have a Write<T> overload that can help validate but that still passes the options along for any nested deserialization.

JsonLogic.Tests/GithubTests.cs
```diff
@@ -12,6 +13,10 @@ namespace Json.Logic.Tests;
 
 public class GithubTests
 {
+#if NET8_0_OR_GREATER
+	private readonly JsonTypeInfo<Rule> _ruleTypeInfo = (JsonTypeInfo<Rule>)Rule.JsonTypeResolver.GetTypeInfo(typeof(Rule), JsonSerializerOptions.Default)!;
```
jevansaks, line 17: Instead of a JsonTypeInfo field, we could make a JsonSerializerOptions and then the below calls can be options.Read<Rule>(jsonRule) which would also remove the `#if` conditions.

JsonLogic/Rule.cs
```diff
@@ -111,13 +121,13 @@ public override Rule Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSer
...
-					? (Rule)JsonSerializer.Deserialize("[]", ruleType, options)!
-					: (Rule)args.Deserialize(ruleType, options)!;
+					? (Rule)JsonSerializer.Deserialize("[]", ruleType, LogicSerializerContext.Default)!
+					: (Rule)args.Deserialize(ruleType, LogicSerializerContext.Default)!;
 			}
 		}
 		else if (node is JsonArray)
 		{
-			var data = node.Deserialize<List<Rule>>(options)!;
+			var data = node.Deserialize(LogicSerializerContext.Default.RuleArray)!;
```
jevansaks, line 130: Can we switch to options.Read here too?

JsonLogic/Rules/LiteralRule.cs
```diff
@@ -48,6 +48,6 @@ internal class LiteralRuleJsonConverter : JsonConverter<LiteralRule>
 
 	public override void Write(Utf8JsonWriter writer, LiteralRule value, JsonSerializerOptions options)
 	{
-		JsonSerializer.Serialize(writer, value.Value, options);
+		JsonSerializer.Serialize(writer, value.Value, LogicSerializerContext.Default.JsonNode!);
```
jevansaks, line 51: Same comment about wanting an options.Write overload here so we get type safety but also preserving options.

JsonLogic/Rules/LooseEqualsRule.cs
```diff
@@ -50,11 +51,14 @@ protected internal LooseEqualsRule(Rule a, Rule b)
 	}
 }
 
-internal class LooseEqualsRuleJsonConverter : JsonConverter<LooseEqualsRule>
+internal class LooseEqualsRuleJsonConverter : AotCompatibleJsonConverter<LooseEqualsRule>
 {
 	public override LooseEqualsRule? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
 	{
-		var parameters = JsonSerializer.Deserialize<Rule[]>(ref reader, options);
+		var readerCopy = reader;
```
jevansaks: Why did this need to change?

JsonLogic/Rules/ReduceRule.cs
```diff
@@ -70,7 +76,9 @@ protected internal ReduceRule(Rule input, Rule rule, Rule initial)
 				Current = element,
 				Accumulator = accumulator
 			};
+#pragma warning disable IL2026, IL3050
 			var item = JsonSerializer.SerializeToNode(intermediary, _options);
```
jevansaks, line 80: Would a '_options.WriteToNode' helper be useful for type safety in cases like this?

jevansaks (changes requested):
I also see that RuleRegistry is using self-reflection and triggering some AOT warnings. Can you update that to a hard-coded list of the rule types? The attribute fetching logic should be fine.

JsonLogic/Rules/LooseEqualsRule.cs
gregsdennis: I was just testing things. Wanted to see what was in there.  Trying to figure out the null issue.

JsonLogic/Rules/ReduceRule.cs
gregsdennis: Yes.  Yes it would.

gregsdennis:
The only thing this needs fixed is the null deserialization issue that apparently STJ v8 fixes.

JsonLogic/Rule.cs
```diff
@@ -111,13 +121,13 @@ public override Rule Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSer
 					throw new JsonException($"Cannot identify rule for {op}");
 
 				rule = args is null
-					? (Rule)JsonSerializer.Deserialize("[]", ruleType, options)!
-					: (Rule)args.Deserialize(ruleType, options)!;
+					? (Rule)JsonSerializer.Deserialize("[]", ruleType, LogicSerializerContext.Default)!
```
jevansaks, line 124: This is technically a behavior change because we're not passing the options along.

I would recommend passing options and suppressing the warning, as unsavory as that sounds.

JsonLogic/RuleRegistry.cs
```diff
@@ -10,12 +16,55 @@ namespace Json.Logic;
...
+	private static readonly ConcurrentDictionary<string, Type> _rules;
+	private static readonly ConcurrentDictionary<Type, JsonSerializerContext> _externalRuleResolvers = new();
+
+#if NET8_0_OR_GREATER
+	internal static IJsonTypeInfoResolver[] ExternalTypeInfoResolvers => _externalRuleResolvers.Values.Distinct().ToArray();
+#endif
+
+	static RuleRegistry()
+	{
+		_rules = new ConcurrentDictionary<string, Type>(new Dictionary<string, Type>
```
jevansaks, line 28: Didn't one of the rules have two operator aliases for it? Eyeballing this I don't see any double types in the list.

jevansaks (approved):
Some comments, but non-blocking if you want to merge this in.

JsonLogic/Rule.cs
gregsdennis: I actually just probably need to use the options extension.  Likely just missed this.

JsonLogic/RuleRegistry.cs
gregsdennis: Yeah, it's the if rule. It's there.

jevansaks:
Here's a workaround for the null issue. Update Read<T> in JsonSerializerOptionsExtensions.cs to include this slightly unsavory workaround:
```
	public static T? Read<T>(this JsonSerializerOptions options, ref Utf8JsonReader reader, JsonTypeInfo<T>? typeInfo = null)
	{
#if !NET8_0_OR_GREATER // Workaround for System.Text.Json 6.x missing fix for https://github.com/dotnet/runtime/issues/85172
		if (reader.TokenType == JsonTokenType.Null && typeof(T) == typeof(JsonNode))
		{
			return default(T);
		}
#endif
		return options.GetConverter<T>(typeInfo).Read(ref reader, typeof(T), options);
	}
```

PR #630, opened 2024-01-28
Retarget netstandard2.0 to System.Text.Json 8.0.1

jevansaks:


jevansaks:
Now that we have a workaround for the null issue, i'll abandon this PR.

PR #631, opened 2024-01-28
Native aot context options mgr

gregsdennis:
This should allow us to use all of those options tricks in the other libs.  (It'll let others use it too!)

### DIFF
--- a/Json.More/TypeResolverOptionsManager.cs
+++ b/Json.More/TypeResolverOptionsManager.cs
@@ -1,0 +1,73 @@
+﻿using System.Text.Encodings.Web;
+using System.Text.Json;
+using System.Text.Json.Serialization.Metadata;
+
+namespace Json.More;
+
+public class TypeResolverOptionsManager
+{
+	private JsonSerializerOptions? _serializerOptions;
+	private JsonSerializerOptions? _serializerOptionsUnsafeRelaxedJsonEscaping;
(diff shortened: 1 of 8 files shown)


gregsdennis:
every time!  ☝️

Json.More/TypeResolverOptionsManager.cs
```diff
@@ -0,0 +1,73 @@
...
+#if NET8_0_OR_GREATER
+	public IJsonTypeInfoResolver TypeInfoResolver => _typeInfoResolver;
+	
+	public TypeResolverOptionsManager(IJsonTypeInfoResolver baseResolver, params IJsonTypeInfoResolver[] resolvers)
+	{
+		_baseResolver = baseResolver;
+		_typeInfoResolver = JsonTypeInfoResolver.Combine([baseResolver, .. resolvers]);
+	}
+
+	public void RebuildTypeResolver(params IJsonTypeInfoResolver[] resolvers)
```
jevansaks, line 63: The reason I had this as "invalidate" was so that we didn't do extra work on each Register call and only built the list the next time someone asked for it. Since the behavior is maintained for the options objects, it's probably ok.

gregsdennis: But the code was also holding onto references outside of itself.

PR #632, opened 2024-01-28
Add options.Write extension method and use it in JsonSchema

jevansaks:


JsonSchema/GlobalSuppressions.cs
gregsdennis, line 1: 🎉

PR #634, opened 2024-01-29
Make tests run in reflection disabled mode

jevansaks:


JsonSchema.Tests/JsonSchema.Tests.csproj
```diff
@@ -10,6 +10,7 @@
 		<SignAssembly>true</SignAssembly>
 		<AssemblyOriginatorKeyFile>../json-everything.snk</AssemblyOriginatorKeyFile>
 		<Nullable>enable</Nullable>
+		<JsonSerializerIsReflectionEnabledByDefault Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">false</JsonSerializerIsReflectionEnabledByDefault>
```
gregsdennis, line 13: I haven't been putting the condition in.  I wonder what the implication is for the lower frameworks.

jevansaks: Only respected by STJ8, which implies net8 for us. But if we ever bump the version for the others, then net6 can still be testing the reflection-enabled flavor.

PR #635, opened 2024-01-29
Get extension libs working with Native AOT and tests passing with reflection disabled

jevansaks:
Requires #634


Json.More/JsonSerializerOptionsExtensions.cs
```diff
@@ -59,23 +59,6 @@ public static void Write<T>(this JsonSerializerOptions options, Utf8JsonWriter w
...
-	/// Write an object to JSON. If the type is known, prefer Write<![CDATA[<T>]]>
-	/// </summary>
-	/// <remarks>
-	/// A converter may throw any Exception, but should throw <cref>JsonException</cref> when the JSON is invalid.
-	/// </remarks>
-	/// <param name="options">The <see cref="JsonSerializerOptions"/> being used.</param>
-	/// <param name="writer">The <see cref="Utf8JsonReader"/> to read from.</param>
-	/// <param name="value">The value to serialize.</param>
-	/// <returns>The value that was converted.</returns>
-	public static void Write(this JsonSerializerOptions options, Utf8JsonWriter writer, object? value)
```
gregsdennis, line 75: Was this not useful?  Do you think it could be useful to users who aren't using AOT?

JsonSchema.ArrayExt/JsonSchema.ArrayExt.csproj
```diff
@@ -42,6 +44,11 @@
 	</ItemGroup>
 
 	<ItemGroup>
+		<PackageReference Include="PolySharp" Version="1.14.1">
+			<PrivateAssets>all</PrivateAssets>
+			<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+		</PackageReference>
+		<PackageReference Include="JetBrains.Annotations" Version="2021.2.0" />
```
gregsdennis, line 51: The Jetbrains should be inherited from JsonSchema.  Maybe you copied an extra line?

JsonSchema.DataGeneration/JsonSchemaExtensions.cs
```diff
@@ -79,8 +80,23 @@ internal static GenerationResult GenerateData(this RequirementsContext requireme
...
+				typeof(EnumRequirementsGatherer),
+				typeof(FalseRequirementsGatherer),
+				typeof(ItemsRequirementsGatherer),
+				typeof(NotRequirementsGatherer),
+				typeof(NumberRequirementsGatherer),
+				typeof(OneOfRequirementsGatherer),
+				typeof(PropertiesRequirementsGatherer),
+				typeof(StringRequirementsGatherer),
+				typeof(TypeRequirementsGatherer),
+			}
```
gregsdennis, line 99: If we're listing out the types anyway, we might as well just instantiate with `new()`.

JsonSchema.OpenApi/JsonSchema.OpenApi.csproj
```diff
@@ -41,6 +43,11 @@
 	</ItemGroup>
 
 	<ItemGroup>
+		<PackageReference Include="PolySharp" Version="1.14.1">
+			<PrivateAssets>all</PrivateAssets>
+			<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+		</PackageReference>
+		<PackageReference Include="JetBrains.Annotations" Version="2021.2.0" />
```
gregsdennis, line 50: Jetbrains copied by accident?

Json.More/JsonSerializerOptionsExtensions.cs
jevansaks: It was no longer called. We could include it with an explicit type parameter. The problem I had was that it broke one of the tests so I updated the caller to go back to Serialize. But as I'm typing this comment I realize it's probably that we didn't pass the explicit type. I can fix that up.

JsonSchema.ArrayExt/JsonSchema.ArrayExt.csproj
jevansaks: Oops.

Json.More/JsonSerializerOptionsExtensions.cs
gregsdennis: Don't worry about it.

JsonSchema.DataGeneration/JsonSchemaExtensions.cs
jevansaks: Oh! I didn't realize that's what this was doing, ha.

JsonSchema.OpenApi/JsonSchema.OpenApi.csproj
jevansaks: Fixed

jevansaks:
Pushed an update that addresses the CR feedback.

PR #636, opened 2024-01-29
Update pragma warning disables to SuppressMessage attributes

jevansaks:
The downstream AOT callers can't see our #pragma warning disables. Need to use attributes so that people compiled against the library don't get messages about things we suppressed.

@sergio0694 can you take a peek too?

### DIFF
--- a/Json.More/JsonSerializerOptionsExtensions.cs
+++ b/Json.More/JsonSerializerOptionsExtensions.cs
@@ -60,9 +60,9 @@ public static class JsonSerializerOptionsExtensions
 	/// <param name="value">The value to serialize.</param>
 	/// <param name="typeInfo">An explicit typeInfo to use for looking up the Converter. If not provided, options.GetTypeInfo will be used.</param>
 	/// <returns>The value that was converted.</returns>
-	[UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "Deserialize is safe in AOT if the JsonSerializerOptions come from the source generator.")]
-	[UnconditionalSuppressMessage("AOT", "IL3050:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "Deserialize is safe in AOT if the JsonSerializerOptions come from the source generator.")]
-	public static void Write<T>(this JsonSerializerOptions options, Utf8JsonWriter writer, T? value, JsonTypeInfo<T>? typeInfo = null)
+	[UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "Serialize is safe in AOT if the JsonSerializerOptions come from the source generator. Requiring the JsonTypeInfo parameter helps enforce that.")]
+	[UnconditionalSuppressMessage("AOT", "IL3050:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "Serialize is safe in AOT if the JsonSerializerOptions come from the source generator. Requiring the JsonTypeInfo parameter helps enforce that.")]
+	public static void Write<T>(this JsonSerializerOptions options, Utf8JsonWriter writer, T? value, JsonTypeInfo<T>? typeInfo)
 	{
 		JsonSerializer.Serialize(writer, value, options);
 	}
@@ -78,10 +78,17 @@ public static class JsonSerializerOptionsExtensions
 	/// <param name="value">The value to serialize.</param>
 	/// <param name="inputType">The type to serialize.</param>
 	/// <returns>The value that was converted.</returns>
-	public static void Write(this JsonSerializerOptions options, Utf8JsonWriter writer, object? value, Type inputType)
+	[RequiresDynamicCode("Calls JsonSerializer.Serialize. Make sure the options object contains all relevant JsonTypeInfos before suppressing this warning.")]
+	[RequiresUnreferencedCode("Calls JsonSerializer.Serialize. Make sure the options object contains all relevant JsonTypeInfos before suppressing this warning.")]
+	public static void Write(this JsonSerializerOptions options, Utf8JsonWriter writer, object? value, Type? inputType)
 	{
-#pragma warning disable IL2026, IL3050 // This helper is expected to be called with an options object that covers the needed TypeInfos.
-		JsonSerializer.Serialize(writer, value, inputType, options);
-#pragma warning restore IL2026, IL3050
+		if (inputType is not null)
+		{
+			JsonSerializer.Serialize(writer, value, inputType, options);
+		}
+		else
+		{
+			JsonSerializer.Serialize(writer, value, options);
+		}
 	}
 }
--- a/Json.More/JsonSerializerOptionsExtensions.cs
+++ b/Json.More/JsonSerializerOptionsExtensions.cs
@@ -80,15 +80,8 @@ public static class JsonSerializerOptionsExtensions
 	/// <returns>The value that was converted.</returns>
 	[RequiresDynamicCode("Calls JsonSerializer.Serialize. Make sure the options object contains all relevant JsonTypeInfos before suppressing this warning.")]
 	[RequiresUnreferencedCode("Calls JsonSerializer.Serialize. Make sure the options object contains all relevant JsonTypeInfos before suppressing this warning.")]
-	public static void Write(this JsonSerializerOptions options, Utf8JsonWriter writer, object? value, Type? inputType)
+	public static void Write(this JsonSerializerOptions options, Utf8JsonWriter writer, object? value, Type inputType)
 	{
-		if (inputType is not null)
-		{
-			JsonSerializer.Serialize(writer, value, inputType, options);
-		}
-		else
-		{
-			JsonSerializer.Serialize(writer, value, options);
-		}
+		JsonSerializer.Serialize(writer, value, inputType, options);
 	}
 }
--- a/Json.More/JsonSerializerOptionsExtensions.cs
+++ b/Json.More/JsonSerializerOptionsExtensions.cs
@@ -64,7 +64,7 @@ public static class JsonSerializerOptionsExtensions
 	[UnconditionalSuppressMessage("AOT", "IL3050:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "Serialize is safe in AOT if the JsonSerializerOptions come from the source generator. Requiring the JsonTypeInfo parameter helps enforce that.")]
 	public static void Write<T>(this JsonSerializerOptions options, Utf8JsonWriter writer, T? value, JsonTypeInfo<T>? typeInfo)
 	{
-		JsonSerializer.Serialize(writer, value, options);
+		options.GetConverter<T?>(typeInfo).Write(writer, value, options);
 	}
 
 	/// <summary>
--- a/Json.More/JsonSerializerOptionsExtensions.cs
+++ b/Json.More/JsonSerializerOptionsExtensions.cs
@@ -60,11 +60,11 @@ public static class JsonSerializerOptionsExtensions
 	/// <param name="value">The value to serialize.</param>
 	/// <param name="typeInfo">An explicit typeInfo to use for looking up the Converter. If not provided, options.GetTypeInfo will be used.</param>
 	/// <returns>The value that was converted.</returns>
+	[UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "Deserialize is safe in AOT if the JsonSerializerOptions come from the source generator.")]
+	[UnconditionalSuppressMessage("AOT", "IL3050:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "Deserialize is safe in AOT if the JsonSerializerOptions come from the source generator.")]
 	public static void Write<T>(this JsonSerializerOptions options, Utf8JsonWriter writer, T? value, JsonTypeInfo<T>? typeInfo = null)
 	{
-#pragma warning disable IL2026, IL3050 // This helper is expected to be called with an options object that covers the needed TypeInfos.
 		JsonSerializer.Serialize(writer, value, options);
-#pragma warning restore IL2026, IL3050
 	}
 
 	/// <summary>
--- a/Json.More/JsonSerializerOptionsExtensions.cs
+++ b/Json.More/JsonSerializerOptionsExtensions.cs
@@ -19,7 +19,7 @@ public static class JsonSerializerOptionsExtensions
 	/// <param name="options">The <see cref="JsonSerializerOptions"/> being used.</param>
 	/// <param name="typeInfo">An explicit typeInfo to use for looking up the Converter. If not provided, options.GetTypeInfo will be used.</param>
 	/// <returns>An implementation of <see cref="JsonConverter{T}"/> as determined by the provided options</returns>
-	public static JsonConverter<T> GetConverter<T>(this JsonSerializerOptions options, JsonTypeInfo? typeInfo = null)
+	public static JsonConverter<T> GetConverter<T>(this JsonSerializerOptions options, JsonTypeInfo? typeInfo)
 	{
 #if NET8_0_OR_GREATER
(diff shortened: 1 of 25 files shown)


Json.More/JsonSerializerOptionsExtensions.cs
```diff
@@ -60,11 +60,11 @@ public static JsonConverter<T> GetConverter<T>(this JsonSerializerOptions option
 	/// <param name="value">The value to serialize.</param>
 	/// <param name="typeInfo">An explicit typeInfo to use for looking up the Converter. If not provided, options.GetTypeInfo will be used.</param>
 	/// <returns>The value that was converted.</returns>
+	[UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "Deserialize is safe in AOT if the JsonSerializerOptions come from the source generator.")]
+	[UnconditionalSuppressMessage("AOT", "IL3050:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "Deserialize is safe in AOT if the JsonSerializerOptions come from the source generator.")]
```
Sergio0694, line 64: I'm not sure about this. Yes the suppression is correct if you are in fact passing options coming from a generated context, but suppressing these warnings in all cases means that callers wanting to use NativeAOT and not using a generated context will not get any warnings at all, but still crash at runtime. This explicitly goes against one of the design principles of the trim/AOT analyzers, we shouldn't hide warnings from users and rely on them doing the right thing (because they won't, and then they'll have a very hard time figuring out what's causing the problems).

JsonSchema.CodeGeneration/JsonSchema.CodeGeneration.csproj
```diff
@@ -52,7 +53,11 @@
 	</ItemGroup>
 
 	<ItemGroup>
-	  <PackageReference Include="Humanizer.Core" Version="2.14.1" />
+		<PackageReference Include="PolySharp" Version="1.14.1">
+			<PrivateAssets>all</PrivateAssets>
+			<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+		</PackageReference>
```
Sergio0694, line 59: Can simplify this:
```suggestion
		<PackageReference Include="PolySharp" Version="1.14.1" PrivateAssets="all" />
```

Json.More/JsonSerializerOptionsExtensions.cs
jevansaks: OK so yes the public methods should carry the RDC&RUC attributes but when we call them internally in a known safe way, we want those suppressed. Is there a good strategy for doing that?

Sergio0694: Can you declare internal proxy methods with those suppression attributes that just call those, and use the proxies?

jevansaks: I see, so the red flag here is that we have a _public_ method annotated with these? But internal methods would be fine?

Sergio0694: Yup! Yeah if they're internal methods and _we_ know _we_ are calling them safely, then it's ok 🙂

jevansaks: I fixed a couple of these, but mostly I just changed the justification message because we're doing the best we can to verify that the options set covers all the types we need. This is how System.Text.Json handles this scenario too.

Json.More/JsonSerializerOptionsExtensions.cs
```diff
@@ -60,11 +60,11 @@ public static JsonConverter<T> GetConverter<T>(this JsonSerializerOptions option
...
 	/// <returns>The value that was converted.</returns>
-	public static void Write<T>(this JsonSerializerOptions options, Utf8JsonWriter writer, T? value, JsonTypeInfo<T>? typeInfo = null)
+	[UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "Serialize is safe in AOT if the JsonSerializerOptions come from the source generator. Requiring the JsonTypeInfo parameter helps enforce that.")]
+	[UnconditionalSuppressMessage("AOT", "IL3050:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "Serialize is safe in AOT if the JsonSerializerOptions come from the source generator. Requiring the JsonTypeInfo parameter helps enforce that.")]
+	public static void Write<T>(this JsonSerializerOptions options, Utf8JsonWriter writer, T? value, JsonTypeInfo<T>? typeInfo)
 	{
-#pragma warning disable IL2026, IL3050 // This helper is expected to be called with an options object that covers the needed TypeInfos.
```
gregsdennis, line 65: If we're not using the type info, why include it as a parameter?

JsonSchema.OpenApi/DiscriminatorKeyword.cs
```diff
@@ -106,13 +107,13 @@ internal class Model
 	/// <param name="typeToConvert">The type to convert.</param>
 	/// <param name="options">An object that specifies serialization options to use.</param>
 	/// <returns>The converted value.</returns>
+	[UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "We guarantee that the SerializerOptions covers all the types we need for AOT scenarios.")]
+	[UnconditionalSuppressMessage("AOT", "IL3050:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "We guarantee that the SerializerOptions covers all the types we need for AOT scenarios.")]
```
gregsdennis, line 111: We should just add `Model` to the context and pass in the type info.

JsonSchema.OpenApi/ExternalDocsKeyword.cs
```diff
@@ -102,13 +103,13 @@ internal class Model
 	/// <param name="typeToConvert">The type to convert.</param>
 	/// <param name="options">An object that specifies serialization options to use.</param>
 	/// <returns>The converted value.</returns>
+	[UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "We guarantee that the SerializerOptions covers all the types we need for AOT scenarios.")]
+	[UnconditionalSuppressMessage("AOT", "IL3050:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "We guarantee that the SerializerOptions covers all the types we need for AOT scenarios.")]
```
gregsdennis, line 107: Add `Model` to the context.

JsonSchema.OpenApi/XmlKeyword.cs
```diff
@@ -141,13 +142,13 @@ internal class Model
 	/// <param name="typeToConvert">The type to convert.</param>
 	/// <param name="options">An object that specifies serialization options to use.</param>
 	/// <returns>The converted value.</returns>
+	[UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "We guarantee that the SerializerOptions covers all the types we need for AOT scenarios.")]
+	[UnconditionalSuppressMessage("AOT", "IL3050:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "We guarantee that the SerializerOptions covers all the types we need for AOT scenarios.")]
```
gregsdennis, line 146: Add `Model` to the context.

Json.More/JsonSerializerOptionsExtensions.cs
jevansaks: For static validation. There seems to be no way to bind a TypeInfo to a different Options than the one JsonSerializerContext that it was born with. Using the TypeInfo ignores any other types in the provided options context, so we prefer the options context.

@eiriktsarpalis this goes to the JsonSerializerContext customization I was asking you for in chat.

gregsdennis: All that's fair, but if we're not using a parameter, why not remove it?

JsonSchema.OpenApi/DiscriminatorKeyword.cs
jevansaks: It's there, what we are missing is SerializeTo/from JsonNode overload to help statically verify this. It just moves the suppression to a different place though, as I mentioned in the other comment -- there's no way to take an options and combine it with a TypeInfo in the STJ methods so this seems the best we can do for now.

JsonSchema.OpenApi/ExternalDocsKeyword.cs
jevansaks: See other comment. The tests wouldn't pass with reflection disabled if it was missing.

Json.More/JsonSerializerOptionsExtensions.cs
gregsdennis: But the `.Read<T>()` method actually uses the parameter to get the appropriate converter.  The `.Write<T>()` method doesn't use the parameter.

Secondly, `.Write<T>()` includes a `T value` parameter that informs the generic type.

gregsdennis: If the `.Write<T>()` method is supposed to mirror the `.Read<T>()` method, I recommend changing the body to

```c#
options.GetConverter<T>(typeInfo).Write(writer, value!, options);
```

jevansaks: Sure, I'd been wanting to do this anyway so let me make that change and see how the tests feel about it.

gregsdennis: (probably remove the `= null` from the parameter, too.)

jevansaks: It works! Updated.

gregsdennis: Do we still need the attributes now?

jevansaks: Good call. Removed.

jevansaks: Actually, wasn't GetConverter<T>() there as a public extension method before? I had the = null as compatibility for external callers. LMK if you think I leave it or bring it back.

gregsdennis: It's not optional on `.Read<T>()`.  I say we keep it off.

Json.More/JsonSerializerOptionsExtensions.cs
```diff
@@ -19,7 +19,7 @@ public static class JsonSerializerOptionsExtensions
 	/// <param name="options">The <see cref="JsonSerializerOptions"/> being used.</param>
 	/// <param name="typeInfo">An explicit typeInfo to use for looking up the Converter. If not provided, options.GetTypeInfo will be used.</param>
 	/// <returns>An implementation of <see cref="JsonConverter{T}"/> as determined by the provided options</returns>
-	public static JsonConverter<T> GetConverter<T>(this JsonSerializerOptions options, JsonTypeInfo? typeInfo = null)
+	public static JsonConverter<T> GetConverter<T>(this JsonSerializerOptions options, JsonTypeInfo? typeInfo)
 	{
 #if NET8_0_OR_GREATER
```
eiriktsarpalis, line 24: It seems that targets older than 8.0 ignore the `typeInfo` argument?

gregsdennis: Yeah, .Net Standard's version doesn't have a `.Converter` property on `JsonTypeInfo`, so we need to just get it off of the options.

Json.More/JsonSerializerOptionsExtensions.cs
eiriktsarpalis: Passing both an `options` and `typeInfo` parameters are redundant and could in fact be mutually contradictory (e.g. because `typeInfo` encapsulates a different options instance). If the motivation is to avoid the linker warning it seems to me that calling
```C#
var typeInfo = (JsonTypeInfo<T>?)options.GetTypeInfo(typeof(T));
if (typeInfo is null) throw new Exception(); /* some error that T is not defined in the options */
return JsonSerializer.Serializer(writer, value, typeInfo);
```

Json.More/JsonSerializerOptionsExtensions.cs
```diff
@@ -78,10 +76,10 @@ public static void Write<T>(this JsonSerializerOptions options, Utf8JsonWriter w
 	/// <param name="value">The value to serialize.</param>
 	/// <param name="inputType">The type to serialize.</param>
 	/// <returns>The value that was converted.</returns>
+	[RequiresDynamicCode("Calls JsonSerializer.Serialize. Make sure the options object contains all relevant JsonTypeInfos before suppressing this warning.")]
+	[RequiresUnreferencedCode("Calls JsonSerializer.Serialize. Make sure the options object contains all relevant JsonTypeInfos before suppressing this warning.")]
```
eiriktsarpalis, line 80: Could be avoided if you called `options.GetTypeInfo()` before passing the result to the  JsonSerializer method accepting JsonTypeInfo

Json.More/JsonSerializerOptionsExtensions.cs
gregsdennis: The goal is that we want to include the type info while also respecting whatever options the user sends us (e.g. write indented, name casing, etc).

Json.More/JsonSerializerOptionsExtensions.cs
gregsdennis: `options.GetTypeInfo()` doesn't exist in .Net Standard 2.0.

Json.More/JsonSerializerOptionsExtensions.cs
eiriktsarpalis: Gotcha. Would it make sense to add a debug assertion?
```C#
Debug.Assert(typeInfo is null || typeInfo.Options == options);
```
Or perhaps even make it a runtime check given that it's public API?

Json.More/JsonSerializerOptionsExtensions.cs
eiriktsarpalis: It should be available to .NET standard 2.0, however that would require referencing a newer version of the STJ NuGet package (>= 7.0.0 for that particular API). I can see why you might want to stay with 6.0.0 as the minimum supported version, but the APIs added in >= 7.0.0 make it more manageable to work in AOT-sensitive components.

gregsdennis: We tried updating to v8 and had issues.  I'm not sure we've tried v7.

eiriktsarpalis: Unless you have customers that specifically require 6.0.0 I would strongly recommend trying to move the minimum version to 8.0.0. It has APIs specifically catering to AOT-ready libraries (e.g. the JsonSerializer.IsReflectionEnabledByDefault property) and it should eliminate the need to multi-target (you can stick with just one netstandard2.0 target)

gregsdennis: Working on it, it seems mostly good now.  Not sure what the errors were before.

There is a discrepancy in double serialization in netcoreapp3.1.  For example `-4.2` is serialized correctly in net6 and net8, but netcoreapp3.1 gives `-4.2000000000000002`.

jevansaks: There weren't really issues, just some "danger use at your own risk" kinds of warnings. I staged a PR with it here but abandoned it: https://github.com/gregsdennis/json-everything/pull/630.

gregsdennis: Oh... There a warning on using STJ v8 with net core 3.1:

> System.Text.Encodings.Web 8.0.0 doesn't support $(TargetFramework) and has not been tested with it. Consider upgrading your TargetFramework to net6.0 or later. You may also set &lt;SuppressTfmSupportBuildWarnings&gt;true&lt;/SuppressTfmSupportBuildWarnings&gt; in the project file to ignore this warning and attempt to run in this unsupported configuration at your own risk.

gregsdennis: Yeah, the 3.1 `double` issue is the only one.  Everything else is working.  I'll start a PR.  Not sure why this was so complicated before.

jevansaks: Great! Once you do that, you should also be able to remove most, if not all, of the NET8_OR_GREATER checks in the code.

gregsdennis: Given that STJ8 doesn't support 3.1 and everything else seems to work, we're dropping 3.1 support as well.  See #640

PR #637, opened 2024-01-29
Get JsonE running in reflection disabled mode

jevansaks:


JsonE/Expressions/ArrayExpressionNode.cs
```diff
@@ -57,7 +58,7 @@ public bool TryParse(ReadOnlySpan<char> source, ref int index, out ExpressionNod
 			if (!ExpressionParser.TryParse(source, ref i, out var value))
 				throw new SyntaxException(CommonErrors.WrongToken(source[i]));
 
-			arr.Add(JsonExpression.Create(value!));
+			arr.Add((JsonNode)(JsonValue)JsonExpression.Create(value!));
```
gregsdennis, line 61: What's going on here?  Double casts are smelly.

JsonE/Operators/JsonOperator.cs
```diff
@@ -36,7 +39,8 @@ internal class JsonOperator : IOperator
 		{
 			dict[kvp.Key] = Sort(kvp.Value);
 		}
-
-		return JsonSerializer.SerializeToNode(dict);
+		
+		// Can't use the TypeInfo overload because JsonSerializerContext differs in nullability
+		return JsonSerializer.SerializeToNode(dict, JsonESerializerContext.Default.Options);
```
gregsdennis, line 44: Is the nullability an error or just a warning?

JsonE/Expressions/ArrayExpressionNode.cs
jevansaks: `arr.Add(JsonNode?)` is AOT safe, but `arr.Add<T>(T)` is not. This makes sure we go through the right one.

I'm open to suggestions of the right way to do this.

JsonE/Operators/JsonOperator.cs
jevansaks: It's this issue I filed:
```suggestion
		// Can't use the TypeInfo overload because JsonSerializerContext differs in nullability (https://github.com/dotnet/runtime/issues/97665)
		return JsonSerializer.SerializeToNode(dict, JsonESerializerContext.Default.Options);
```

JsonE/Expressions/ArrayExpressionNode.cs
gregsdennis: Oh, I didn't even notice it was using a generic!

Maybe just put it in a variable first?
```suggestion
			JsonNode expr = JsonExpression.Create(value!);
			arr.Add(expr);
```

How does that sit?

jevansaks: Yeah, if that compiles I'll take it. :)

PR #638, opened 2024-01-29
update version and release notes

gregsdennis:
Updates all of the versions and release notes.

PR #640, opened 2024-01-30
Native aot stj8

gregsdennis:
Still has issues with `double` in netcoreapp3.1.  Not sure what to do there.

Also supposedly we can stop targeting net8 directly and just use netstandard 2.  I'll try that.


gregsdennis:
I think I'm going to keep the multitarget in the project files.  There are still a few things (like `Enum.GetValues<T>()`) that require it.  Also, we get the compiler warnings only when devving in .net 8.

jevansaks:
Multi-targeting also enables us to run tests reflection enabled and disabled, so that's also handy.

Json.More/Json.More.csproj
```diff
@@ -34,7 +34,7 @@
   </PropertyGroup>
 
   <ItemGroup Condition="'$(TargetFramework)'=='netstandard2.0'">
-    <PackageReference Include="System.Text.Json" Version="6.0.0" />
+    <PackageReference Include="System.Text.Json" Version="8.0.0" />
   </ItemGroup>
```
jevansaks, line 38: Latest is 8.0.1, can we just specify it without any Condition so we know we're on the same version across all flavors?

gregsdennis: Well... is it really necessary to specify it at all in .Net 8?  If you're not in .Net 8, you get the .Net Standard 2.0 lib.  But if you're _in_ .Net 8, then the package is built in, and the reference isn't necessary.

JsonLogic/JsonWriterExtensions.cs
```diff
@@ -22,19 +23,21 @@ public static void WriteRule(this Utf8JsonWriter writer, Rule? rule, JsonSeriali
 			writer.WriteNullValue();
 			return;
 		}
+
 #pragma warning disable IL2026, IL3050
+		// using the options.Write() extension here for some reason causes a StackOverflow
```
jevansaks, line 28: >// using the options.Write() extension here for some reason causes a StackOverflow [](http://example.com/codeflow?start=2&length=82)

My psychic debugging says that you might have dropped the rule.GetType() part and we landed in options.Write<Rule>(...) which then circled back to the same converter we're in.

Use the options.Write(writer, object, Type) overload to navigate yourself into the correct serializer.

jevansaks:
> 

Having done this a bunch these last few days, I prefer the in-source suppressions are better for reviewing/maintaining the code.

---
Refers to: JsonSchema.Generation/GlobalSuppressions.cs:18 in fb79d00. [](commit_id = fb79d00831e5432ae280f520cbd6ee5dba7f473d, deletion_comment = False)

JsonLogic/JsonWriterExtensions.cs
gregsdennis: I'll try that.  Still needs the warnings, though.

Json.More/Json.More.csproj
jevansaks: But in NET8 which version are we getting? Is it always latest? I suspect it's tied to whatever version of the SDK we compiled against. Having one be explicitly specified and the other implicit seems like a recipe for getting inconsistent versions between net6 and net8 and getting bugs that'll make you tear your hair out.

gregsdennis: (I'm just addressing the build condition.  I'm happy with updating the reference to current latest.)

There's nothing guaranteeing that this library will remain updated with the latest either.  If you bump your local framework to 8.0.3, but this lib hasn't updated yet, it'll still use the 8.0.1 from the package.  I'd prefer it use the local framework.  If the local framework is before 8.0.0, then they'll get the .Net Standard version of this lib along with whatever stj8 version this lib needs.

JsonLogic/JsonWriterExtensions.cs
gregsdennis: Nope.  That still causes the error.

jevansaks: You can check in with this workaround but I don't like unsolved mysteries so I will try to debug this tomorrow.

jevansaks: Which test trips the stack overflow? I checked out this branch locally, changed to `options.Write(writer, rule, rule.GetType());` and all the tests pass. 🤔

gregsdennis: Did you run the entire suite?

jevansaks: I thought so. Let me try again.

jevansaks: Same result: 
image

jevansaks: I looked at the file history and the previous was:
```
options.Write(writer, rule, LogicSerializerContext.Default.Rule);
```
Which, because it's being passed Rule and not the specific derived rule that we have in hand, we end up back in the base Rule's RuleConverter.

What I'm proposing is:
```
		options.Write(writer, rule, rule.GetType());
```
Which, yes, requires a suppression on the method, but is needed because we have to get to the derived type's converter and we don't know it without calling GetType().

gregsdennis: Okay, yeah, I swear I tried that...  I had to restart VS, though.  Maybe something went... off.

gregsdennis:
> I prefer the in-source suppressions are better for reviewing/maintaining the code.

I think this'll go in another PR.

JsonLogic/JsonWriterExtensions.cs
```diff
@@ -25,8 +25,7 @@ public static void WriteRule(this Utf8JsonWriter writer, Rule? rule, JsonSeriali
 		}
 
 #pragma warning disable IL2026, IL3050
-		// using the options.Write() extension here for some reason causes a StackOverflow
-		JsonSerializer.Serialize(writer, rule, rule.GetType(), LogicSerializerContext.OptionsManager.SerializerOptions);
+		options.Write(writer, rule, rule.GetType());
```
jevansaks, line 28: FYI still need to use Suppression attributes instead of `#pragma warning disable` because library consumers don't see the `#pragma warning` and the callers will get AOT warnings on our code. The AOT analyzer only sees attributes so we need to suppress that way. See #636

gregsdennis: Added it in #638

PR #642, opened 2024-01-30
Fix ErrorMessages and EvaluationResult.Fail to be AOT compatible

jevansaks:
Mark `ErrorMessages.ReplaceTokens` and `EvaluationResult.Fail` as AOT incompatible. Added overloads for callers to pass JsonTypeInfos, as well as convenience overloads for primitive types that JsonSchemaSerializerContext supports.


JsonSchema/ErrorMessages.cs
```diff
@@ -51,27 +43,63 @@ private static string Get(CultureInfo? culture = null, [CallerMemberName] string
...
+
+	/// <summary>
+	/// Helper to make a parameter for ReplaceTokens in a with static type checking.
+	/// </summary>
+	/// <typeparam name="T"></typeparam>
+	/// <param name="token"></param>
+	/// <param name="value"></param>
+	/// <param name="typeInfo"></param>
+	/// <returns></returns>
+	public static (string token, object? value, JsonTypeInfo typeInfo) MakeParam<T>(
```
gregsdennis, line 100: I feel this can still be cleaner.

What if we create `(string token, T value, JsonTypeInfo<T> typeInfo)` as an actual type (e.g. `ErrorMessageToken<T>` with base `ErrorMessageToken`), then we define implicit operators from the tuple to the type?

Then, we update `.ReplaceTokens()` to take `params ErrorMessageToken[]`.  This method should be able to go away.

(I think `ErrorMessageToken` could also be a child class of `ErrorMessages`, so maybe just `Token`?)

jevansaks: I went down that path for a bit but with parameter lists it just wasn't matching / hitting the implicit operators I wanted it to. The version here is the best I could come up with. Perhaps you can check out the branch and try playing around with it to see if you can get it to work?

gregsdennis: I'll have a play with it in a bit.

gregsdennis: After playing with it, trying to get the implicit cast working (it didn't), I landed on chaining together multiple replacement calls.

```c#
/// <summary>
/// Replaces tokens in the form of `[[token]]` with a specified value, serialized as JSON.
/// </summary>
/// <param name="message">The message template.</param>
/// <param name="name">The token name (without brackets)</param>
/// <param name="value">The value.</param>
/// <param name="typeInfo">A JsonTypeInfo that matches <paramref name="value"/>.</param>
/// <returns>The detokenized string.</returns>
public static string ReplaceToken<T>(this string message, string name, T? value, JsonTypeInfo<T> typeInfo)
{
	var stringValue = JsonSerializer.Serialize(value, typeInfo!);
	return message.Replace($"[[{name}]]", stringValue);
}
```

Usage looks like this:

```c#
evaluation.Results.Fail(Name, "Item at index [[index]] does not have a value at [[pointer]]"
    .ReplaceToken("index", i - 1, ArrayExtSerializerContext.Default.Int32)
    .ReplaceToken("pointer", specifier.By, ArrayExtSerializerContext.Default.JsonPointer)
);
```

jevansaks: ReplaceTokens and the params overload of Fail are gone, ReplaceToken is the replacement. Take a look and see what you think.

JsonSchema/ErrorMessages.cs
```diff
@@ -51,27 +45,82 @@ private static string Get(CultureInfo? culture = null, [CallerMemberName] string
...
+	}
+
+	/// <summary>
+	/// Replaces tokens in the form of `[[token]]` with a specified value, serialized as JSON.
+	/// </summary>
+	/// <param name="message">The message template.</param>
+	/// <param name="name">The token name (without brackets)</param>
+	/// <param name="value">The value.</param>
+	/// <returns>The detokenized string.</returns>
+	public static string ReplaceToken(this string message, string name, int value) =>
```
gregsdennis, line 84: Now that you've done these, I suppose we'd keep them, but I would like one for `JsonNode` to allow users to pre-serialize values.

JsonSchema/FormatKeyword.cs
```diff
@@ -99,9 +99,9 @@ private void AssertionEvaluator(KeywordEvaluation evaluation, EvaluationContext
 		if (Value is UnknownFormat)
 			evaluation.Results.Fail(Name, errorMessage);
 		else if (errorMessage == null)
-			evaluation.Results.Fail(Name, ErrorMessages.GetFormat(context.Options.Culture), ("format", Value.Key));
+			evaluation.Results.Fail(Name, ErrorMessages.GetFormat(context.Options.Culture).ReplaceToken("format", Value.Key));
 		else
-			evaluation.Results.Fail(Name, ErrorMessages.GetFormatWithDetail(context.Options.Culture), ("format", Value.Key), ("detail", errorMessage));
+			evaluation.Results.Fail(Name, ErrorMessages.GetFormatWithDetail(context.Options.Culture).ReplaceToken("format", Value.Key).ReplaceToken("detail", errorMessage));
```
gregsdennis, line 104: Please wrap these on `.ReplaceTokens()`.

JsonSchema/JsonSchemaSerializerContext.cs
```diff
@@ -79,6 +79,7 @@ namespace Json.Schema;
 [JsonSerializable(typeof(Dictionary<string, PropertyDependency>))]
 [JsonSerializable(typeof(Dictionary<string, SchemaOrPropertyList>))]
 [JsonSerializable(typeof(Dictionary<string, string[]>))]
+[JsonSerializable(typeof(IReadOnlyCollection<JsonNode?>))]
```
gregsdennis, line 82: Pretty sure this was already there.  (See line 90.)

JsonSchema/EnumKeyword.cs
```diff
@@ -94,7 +94,9 @@ public EnumKeyword(IEnumerable<JsonNode?> values)
 	private void Evaluator(KeywordEvaluation evaluation, EvaluationContext context)
 	{
 		if (!Values.Contains(evaluation.LocalInstance, JsonNodeEqualityComparer.Instance))
-			evaluation.Results.Fail(Name, ErrorMessages.GetEnum(context.Options.Culture), ("received", evaluation.LocalInstance), ("values", Values));
+			evaluation.Results.Fail(Name, ErrorMessages.GetEnum(context.Options.Culture).
+				ReplaceToken("received", evaluation.LocalInstance, JsonSchemaSerializerContext.Default.JsonNode!).
+				ReplaceToken("values", Values, JsonSchemaSerializerContext.Default.IReadOnlyCollectionJsonNode!));
```
gregsdennis, line 99:
```suggestion
			evaluation.Results.Fail(Name, ErrorMessages.GetEnum(context.Options.Culture)
				.ReplaceToken("received", evaluation.LocalInstance, JsonSchemaSerializerContext.Default.JsonNode!)
				.ReplaceToken("values", Values, JsonSchemaSerializerContext.Default.IReadOnlyCollectionJsonNode!));
```

Please follow this pattern.

JsonSchema/ErrorMessages.cs
jevansaks: Having the strong overloads is a life saver, IMO. I just added a JsonNode version.

JsonSchema/EnumKeyword.cs
jevansaks: Fixed everywhere

JsonSchema/ErrorMessages.cs
gregsdennis: I imagine it gives a sufficient example to form a pattern for others who want to make their own strongly typed ones as well.

PR #645, opened 2024-01-31
update to expose contexts

gregsdennis:
Really just want to show the Logic lib.  The rest isn't quite ready.

This showcases that we can expose the contexts, which gives the user access to the JTIs for all of the types contained in or used by the lib.  We still use the options manager to handle deserializing user-registered rules, but it's kept internal.  The hard part is remembering on which context your JTI exists.

Also, because the context is exposed, there's no need to have the `IJsonTypeInfoResolver` static property.

Ultimately, I see this working out for all the libs.  We can also create a serializer context in Json.More that exposes common types, like JsonNode.  (Really, I think .Net should expose a context for its common types... \*cough\* @eiriktsarpalis)

We might also be able to drop the `IJsonConverterReadWrite`.


JsonSchema.OpenApi/DiscriminatorKeyword.cs
```diff
@@ -55,7 +55,7 @@ public DiscriminatorKeyword(string propertyName, IReadOnlyDictionary<string, str
 		Mapping = mapping;
 		Extensions = extensions;
 
-		_json = JsonSerializer.SerializeToNode(this, OpenApiSerializerContext.OptionsManager.SerializerOptions);
+		_json = JsonSerializer.SerializeToNode(this, OpenApiSerializerContext.Default.JsonNode);
```
jevansaks, line 58: Why is this JsonNode and not DiscriminatorKeyword? This seems like a mismatch

gregsdennis: oh... ignore that.  Just look at the logic stuff.  That's the only thing guaranteed to work so far.

JsonLogic/Rule.cs
```diff
@@ -119,9 +111,11 @@ public override Rule Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSer
 				if (ruleType == null)
 					throw new JsonException($"Cannot identify rule for {op}");
 
+				var typeInfo = LogicSerializerContext.OptionsManager.SerializerOptions.GetTypeInfo(ruleType);
```
jevansaks, line 114: Let's just go straight to the RuleRegistry's lookup table and then we don't need to go through the ExternalTypeInfoResolvers thing. We might even be able to get rid of the OptionsManager.

gregsdennis: I'll give that a go.

jevansaks:
> 		where T : Rule

Now that we are targeting latest System.Text.Json, we can change this parameter on `AddRule<T>` and `RegisterKeyword<T>`  to be an IJsonTypeInfoResolver. I had only made it a JsonSerializerContext because STJ6 didn't have the interface.

---
Refers to: JsonLogic/RuleRegistry.cs:118 in 075c705. [](commit_id = 075c7050a4cc907bf886a955a13a8fe3e2bdbcee, deletion_comment = False)

JsonSchema/SchemaKeywordRegistry.cs
```diff
@@ -102,6 +102,7 @@ static SchemaKeywordRegistry()
 		};
 
 		_keywords = new ConcurrentDictionary<string, Type>(keywordData.ToDictionary(x => x.Item2, x => x.Item1));
+		_keywordTypeInfoResolvers = new ConcurrentDictionary<Type, JsonSerializerContext>(keywordData.ToDictionary(x => x.Item1, _ => (JsonSerializerContext)JsonSchemaSerializerContext.Default));
```
jevansaks, line 106: RuleRegistry did this differently and just returns null for known types. It seems more efficient to not populate this dictionary and just return the default context when the keyword type isn't found.

JsonSchema/SchemaKeywordRegistry.cs
```diff
@@ -173,6 +172,13 @@ public static void Unregister<T>()
 			: null;
 	}
 
+	internal static JsonTypeInfo GetTypeInfo(Type ruleType)
+	{
+		return _keywordTypeInfoResolvers.TryGetValue(ruleType, out var context)
+			? context.GetTypeInfo(ruleType)!
+			: JsonSchemaSerializerContext.Default.GetTypeInfo(typeof(UnrecognizedKeyword))!;
+	}
```
jevansaks, line 180: Why is this ignoring ruleType here and passing UnrecognizedKeyword explicitly?

JsonSchema.Data/Vocabularies.cs
```diff
@@ -31,27 +30,29 @@ public static void Register(VocabularyRegistry? vocabRegistry = null, SchemaRegi
...
+	internal static JsonSerializerOptions CombinedOptions { get; }
 
-	static DataExtSerializerContext() {
-		OptionsManager = new(
-			Default,
-			JsonSchema.TypeInfoResolver
-		);
+	static JsonSchemaDataSerializerContext()
+	{
+		CombinedOptions = new JsonSerializerOptions
```
jevansaks, line 53: Why do we need to combine options anymore? This new strategy should avoid the need for it.

JsonSchema.Data.Tests/Suite/Suite.cs
```diff
@@ -43,10 +55,7 @@ public static IEnumerable<TestCaseData> TestCases()
 		{
 			var shortFileName = System.IO.Path.GetFileNameWithoutExtension(fileName);
 			var contents = File.ReadAllText(fileName);
-			var collections = JsonSerializer.Deserialize<List<TestCollection>>(contents, new JsonSerializerOptions(DataTestsSerializerContext.OptionsManager.SerializerOptions)
-			{
-				PropertyNameCaseInsensitive = true
-			});
+			var collections = JsonSerializer.Deserialize<List<TestCollection>>(contents, _serializerOptions);
```
jevansaks, line 58: Similarly, this call should be able to just pass DataTestsSerializerContext.Default.ListTestCollection, right?

Json.More/AotCompatibleJsonConverter.cs
```diff
@@ -21,7 +22,8 @@ public interface IJsonConverterReadWrite
 		/// <param name="writer">The writer to write to.</param>
 		/// <param name="value">The value to convert to JSON.</param>
 		/// <param name="options">An object that specifies serialization options to use.</param>
-		void Write(Utf8JsonWriter writer, object value, JsonSerializerOptions options);
+		/// <param name="typeInfo">The <see cref="JsonTypeInfo"/> for the value being serialized.</param>
+		void Write(Utf8JsonWriter writer, object value, JsonSerializerOptions options, JsonTypeInfo? typeInfo = null);
```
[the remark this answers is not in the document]
gregsdennis: Read is working.  I didn't need to update it.  I can look at doing that, though.

JsonSchema/SchemaKeywordRegistry.cs
gregsdennis: If the keyword isn't found, it's treated as an unrecognized keyword.  This captures the data for the keyword.  It doesn't play into validation and just generates and annotation.

JsonSchema.Data/Vocabularies.cs
gregsdennis: This was the workaround for the bug.  I need it in the `UriIdentifier` class where I'm serializing a `JsonSchema`.

JsonSchema.Data.Tests/Suite/Suite.cs
gregsdennis: I wanted to be able to set the options here.

JsonSchema.Data/Vocabularies.cs
jevansaks: Which bug? Can you include a comment / pointer to the issue if it's a bug in STJ?

gregsdennis:
closing in favor of #648 which includes these changes.

PR #648, opened 2024-02-02
Native aot with ~options builder methods~ just the right attributes on the test contexts

gregsdennis:
### Update

(see also "original intent" below)

This PR updates all of the serializer contexts back to internal while ensuring that all of the types are supported.

Additionally, it creates several helper methods to work around a [known STJ bug](https://github.com/dotnet/runtime/issues/50205).  These methods are used throughout the libraries to ensure that the options passed in by the user are respected while still being able to use the serializer context JTIs, which are needed for AOT.

The end result is that contexts are hidden, and users don't need to have annotations for all of the types contained within these libraries; only those which they serialize directly.  For example, for JsonSchema, this will likely be only `JsonSchema` and `EvaluationResults`.

---

### Original intent of the PR

This implements "sharing" the contexts via `JsonSerializerOptions` extensions that simply tack on the context to the options' `.TypeInfoResolverChain`.

There is a single bug that I'm experiencing, and I don't know why.  Specifically `JsonPointer` won't serialize inside the `Json.Schema.EvaluationResultsConverter.Write()` method.  When attempting to [serialize the `.Details` collection](https://github.com/gregsdennis/json-everything/pull/648/files#diff-5b37e9cc329e0c0e98c574a9ffc547931c0199bf198dd91a6930595b9ee4b4a1L355) (which is `IReadOnlyList<EvaluationResults>`), I pass in the options, and I have confirmed that at this point all of the resolvers are present in the options.  Later, through nested serialization, it comes back into the `.Write()` method with the `JsonSchemaSerializerContext.Default.Options` only, which means that it's lost all of the type resolvers somehow, and it fails on `JsonPointer`.

Now, I know that I can just add `JsonPointer` to the internal serializer context, but I'd really like to figure out why the options object is changing as this is the root cause.


JsonSchema/EvaluationResults.cs
```diff
@@ -311,7 +313,7 @@ public override void Write(Utf8JsonWriter writer, EvaluationResults value, JsonS
 		if (value.Format == OutputFormat.Hierarchical || value.Parent != null)
 		{
 			writer.WritePropertyName("evaluationPath");
-			options.Write(writer, value.EvaluationPath, JsonSchemaSerializerContext.Default.JsonPointer);
+			JsonSerializer.Serialize(writer, value.EvaluationPath, options);
```
jevansaks, line 316: I don't understand why these needed to change. The way this options.Write pattern works is that it takes the TypeInfo to get directly to the Converter and calls its Write method, passing the options along. @eiriktsarpalis recommended that approach earlier, and I like that it statically checks that we have the JsonTypeInfos that we need in our own converters.

gregsdennis: Because I removed JsonPointer from the JsonSchema context. (I've updated the description.)

gregsdennis:
I did notice that the serialization after going into the `options.Write` for the `.Details` collection, it ends up at the "AOT-aware users should not get this far" [line](https://github.com/gregsdennis/json-everything/pull/648/files#diff-b77e282aa0bc798ae8a2896d26f747c40b05238b19fe774bb4201fe7bb712fd2R69) because we don't have an explicit aot-compatible-converter for the collection.

The options still has the full resolver chain at this point.  Whatever is happening, it's in the external code between this line and going back into the `EvaluationsResultsConverter`.

JsonSchema/EvaluationResults.cs
jevansaks: I don't think we should do that though, because EvaluationResultsJsonConverter needs to serialize a JsonPointer type, so this component should guarantee that it has a JsonTypeInfo for a JsonPointer instead of relying on the callers to chain them together into the options.

JsonPatch/PatchOperation.cs
```diff
@@ -203,34 +206,34 @@ public override void Write(Utf8JsonWriter writer, PatchOperation value, JsonSeri
 		writer.WriteStartObject();
 
 		writer.WritePropertyName("op");
-		JsonSerializer.Serialize(writer, value.Op, PatchSerializerContext.Default.OperationType);
+		JsonSerializer.Serialize(writer, value.Op, JsonPatchSerializerContext.Default.OperationType);
 
 		writer.WritePropertyName("path");
-		JsonSerializer.Serialize(writer, value.Path, PatchSerializerContext.Default.JsonPointer);
+		JsonSerializer.Serialize(writer, value.Path, options);
```
jevansaks, line 212: This should go back to passing PatchSerializerContext.Default.JsonPointer

JsonPath/JsonSerializerOptionsExtensions.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Json.Path;
+
+public static class JsonSerializerOptionsExtensions
+{
+	public static JsonSerializerOptions WithJsonPath(this JsonSerializerOptions options)
```
jevansaks, line 8: I don't think this is necessary, since JsonPath library doesn't have any types it needs for its converters to succeed. The pattern is needed if we have TypeInfos we need to dynamically look up in the TypeInfo dictionary to serialize types in this library, but there aren't any.

gregsdennis: I may not be, but a user might.  Would this not help them?

jevansaks: If they to serialize a JsonPath they should do [JsonSerializable(JsonPath)] on their own JsonSerializerContext. The pattern is: if you want to serialize those types, you should generate a JsonTypeInfo for them in your own assembly and be self-complete.

JsonSchema.Generation.Tests/TestEnvironment.cs
```diff
@@ -1,24 +1,23 @@
...
-	{
-		
-	}
 }
+
+[JsonSerializable(typeof(JsonSchema))]
+[JsonSerializable(typeof(PropertiesKeyword))]
+[JsonSerializable(typeof(int))]
+[JsonSerializable(typeof(string))]
+[JsonSerializable(typeof(bool))]
```
gregsdennis, line 22: Not sure why I had to add these.  They're all used only internally to serializing a `JsonNode`.  Adding `JsonNode` here didn't help.

JsonSchema.OpenApi/JsonSchemaOpenApiSerializerContext.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Text.Json.Serialization;
+
+namespace Json.Schema.OpenApi;
+
+[JsonSerializable(typeof(ExampleKeyword))]
+[JsonSerializable(typeof(DiscriminatorKeyword))]
+[JsonSerializable(typeof(ExternalDocsKeyword))]
+[JsonSerializable(typeof(XmlKeyword))]
+[JsonSerializable(typeof(JsonNode))]
+[JsonSerializable(typeof(IReadOnlyDictionary<string, string>))]
```
gregsdennis, line 12: Also not sure why I had to add this.

JsonLogic/Rules/AllRule.cs
```diff
@@ -57,11 +57,11 @@ protected internal AllRule(Rule input, Rule rule)
 	}
 }
 
-internal class AllRuleJsonConverter : AotCompatibleJsonConverter<AllRule>
+internal class AllRuleJsonConverter : WeaklyTypedJsonConverter<AllRule>
 {
 	public override AllRule? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
 	{
-		var parameters = options.Read(ref reader, LogicSerializerContext.Default.RuleArray);
+		var parameters = options.Read(ref reader, JsonLogicSerializerContext.Default.RuleArray);
```
jevansaks, line 65: These should switch to options.ReadArray. It's probably fine because we don't have any tests that are trying to pass in an options dictionary that lacks the Rule type so we aren't tripping the bug in all these rule converters.

JsonLogic/JsonSerializerOptionsExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+[JsonSerializable(typeof(RuleCollection))]
+[JsonSerializable(typeof(SomeRule))]
+[JsonSerializable(typeof(StrictEqualsRule))]
+[JsonSerializable(typeof(StrictNotEqualsRule))]
+[JsonSerializable(typeof(SubstrRule))]
+[JsonSerializable(typeof(SubtractRule))]
+[JsonSerializable(typeof(VariableRule))]
+[JsonSerializable(typeof(JsonNode))]
+[JsonSerializable(typeof(Rule[]))]
+[JsonSerializable(typeof(int))]
```
jevansaks, line 45: Recommend removing this so we don't accidentally use it and trigger the STJ bug

JsonPatch/JsonPatch.cs
```diff
@@ -103,7 +97,7 @@ public class PatchJsonConverter : AotCompatibleJsonConverter<JsonPatch>
 	/// <returns>The converted value.</returns>
 	public override JsonPatch Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
 	{
-		var operations = JsonSerializer.Deserialize(ref reader, PatchSerializerContext.Default.ListPatchOperation)!;
+		var operations = JsonSerializer.Deserialize(ref reader, JsonPatchSerializerContext.Default.ListPatchOperation)!;
```
jevansaks, line 101: Switch these to options.ReadList / options.WriteList?

JsonPatch/JsonPatchSerializerContext.cs
```diff
@@ -0,0 +1,19 @@
...
+/// A serializer context for this library.
+/// </summary>
+[JsonSerializable(typeof(JsonPatch))]
+[JsonSerializable(typeof(PatchOperation))]
+[JsonSerializable(typeof(OperationType))]
+[JsonSerializable(typeof(PatchResult))]
+[JsonSerializable(typeof(JsonNode))]
+[JsonSerializable(typeof(JsonPointer))]
+[JsonSerializable(typeof(List<PatchOperation>))]
+[JsonSerializable(typeof(IReadOnlyList<PatchOperation>))]
```
jevansaks, line 18: Recommend removing the collection types

JsonPatch/PatchOperation.cs
```diff
@@ -161,9 +161,12 @@ internal class Model
 		public JsonElement Value { get; set; }
 	}
 
+	[JsonSerializable(typeof(Model))]
+	internal partial class InternalPatchOperationContext : JsonSerializerContext;
+
 	public override PatchOperation Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
 	{
-		var model = JsonSerializer.Deserialize(ref reader, PatchSerializerContext.Default.Model)!;
+		var model = JsonSerializer.Deserialize(ref reader, InternalPatchOperationContext.Default.Model)!;
```
jevansaks, line 170: should these use options.Read? If they were using JsonSerializer.Deserialize in the master branch then it's probably fine to keep.

JsonSchema/JsonSchemaSerializerContext.cs
```diff
@@ -73,11 +75,9 @@ namespace Json.Schema;
...
 [JsonSerializable(typeof(WriteOnlyKeyword))]
 [JsonSerializable(typeof(decimal))]
+[JsonSerializable(typeof(SchemaOrPropertyList))]
 [JsonSerializable(typeof(Dictionary<string, bool>))]
-[JsonSerializable(typeof(Dictionary<string, JsonSchema>))]
-[JsonSerializable(typeof(Dictionary<string, List<string>>))]
 [JsonSerializable(typeof(Dictionary<string, PropertyDependency>))]
-[JsonSerializable(typeof(Dictionary<string, SchemaOrPropertyList>))]
 [JsonSerializable(typeof(Dictionary<string, string[]>))]
 [JsonSerializable(typeof(IReadOnlyCollection<JsonNode?>))]
```
jevansaks, line 82: I didn't finish this last night, but we should remove all the collection types off here just to be safe.

JsonSchema.Generation.Tests/TestEnvironment.cs
jevansaks: @eiriktsarpalis ? I suspect that JsonNode's converter looks up TypeInfos in the options dictionary even for these POD types. @gregsdennis if you can share a full call stack (turning off just my code) for when we fail without these, that would be helpful for Eirik to take a look at.

Json.More/JsonSerializerOptionsExtensions.cs
```diff
@@ -51,9 +52,204 @@ public static JsonConverter<T> GetConverter<T>(this JsonSerializerOptions option
 	/// <param name="value">The value to serialize.</param>
 	/// <param name="typeInfo">An explicit typeInfo to use for looking up the Converter. If not provided, options.GetTypeInfo will be used.</param>
 	/// <returns>The value that was converted.</returns>
-	public static void Write<T>(this JsonSerializerOptions options, Utf8JsonWriter writer, T? value, JsonTypeInfo<T>? typeInfo)
+	public static void Write<T>(this JsonSerializerOptions options, Utf8JsonWriter writer, T value, JsonTypeInfo<T> typeInfo)
```
eiriktsarpalis, line 55: Not sure if I mentioned this in another PR, but the `options` and `typeInfo` parameters defined here are redundant and mixing them together can result in consistency problems. Either resolve the `typeInfo` from the `options.GetTypeInfo` method or get the `Options` instance from the `typeInfo` parameter.

JsonLogic.Tests/GithubTests.cs
```diff
@@ -135,7 +135,7 @@ public void Issue313_VarEqualsNull()
   ]
 }");
 
-		var rule = node.Deserialize<Rule>(LogicSerializerContext.OptionsManager.SerializerOptions);
+		var rule = node.Deserialize(JsonLogicSerializerContext.Default.Rule);
```
eiriktsarpalis, line 138: Not sure if the project structure allows this, but creating separate contexts for each logical grouping of types is bound to create _some_ duplication (e.g. each grouping emits code for common JsonTypeInfo types such as string). If possible, it might make sense to pile as many types as possible in a single context so that common transitive dependencies are shared.

Json.More/JsonSerializerOptionsExtensions.cs
jevansaks: The TypeInfo here is used just to get the Converter, and we're using a JsonTypeInfo as parameter to make sure the caller has tagged the type on a JsonSerializerContext. The Options dictionary is used for all caller-defined types and any other configurations.

JsonSchema.Generation.Tests/TestEnvironment.cs
eiriktsarpalis: Yeah, it's one of the unfortunate issues with `JsonNode` design. It has been designed to encapsulate arbitrary .NET values so depending on how the `JsonNode` is constructed it can be representing JSON numbers or strings as `JsonValue<int>` and `JsonValue<string>`. From that point on, there's no way it can serialize itself without being able to obtain a `JsonTypeInfo<T>` from the underlying `JsonSerializerOptions`.

If this has been caused by something like
```C#
new JsonArray { 42, "string", false }
```
you could try to replace it either with
```C#
new JsonArray { (JsonNode)42, (JsonNode)"string", (JsonNode)false };
```
or just
```C#
JsonNode.Parse("""[42, "string", false]""")
```

gregsdennis: Oh, yeah, we changed a bunch of array initializers with ctor calls before. I'll have a scan for any that we missed.

Json.More/JsonSerializerOptionsExtensions.cs
gregsdennis: (I'm still first on why the your info is tied to an options in the first place. They seem to be independent concepts.)

JsonLogic.Tests/GithubTests.cs
gregsdennis: Oh, yeah that context is internal, but the test project can see the internals. I need to scan for context usages like that and not do it.  Good catch.

JsonSchema.Generation.Tests/TestEnvironment.cs
gregsdennis: (That kinda defeats the purpose of the implicit cast, though, doesn't it?)

jevansaks: I debugged this a little, the problem is a little subtle. It's that the JsonSchema.Generation library is calling JsonNode.Create(object) and then JsonSchema tries to serialize that later. Because the options came from a JsonSerializerContext, we only look for those POD types out of that options dictionary. We could try to call a helper instead of JsonValue.Create everywhere in our code, like this (below). I tried this locally and it seems to address the problem and lets me remove the primitive types from the client JsonSerializerContext.
```
	public static JsonValue? Create(object? value)
	{
		if (value is bool b)
		{
			return JsonValue.Create(b);
		}
		else if (value is double d)
		{
			return JsonValue.Create(d);
		}
		else if (value is int i)
		{
			return JsonValue.Create(i);
		}
		else if (value is string s)
		{
			return JsonValue.Create(s);
		}
		else
		{
			return JsonValue.Create(value);
		}
	}
```

jevansaks: (And if you go this path, we should add in the missing POD types that JsonValue supports as builtin. (this list is missing char and float, etc)

gregsdennis: Looks like that call is just in the `ConstIntent` and `DefaultIntent`, which actually comes from the associated attributes.

I'd like to:

- change the intent class ctors to take `JsonNode` instead, since that's what the `ConstKeyword` takes
- create strongly typed overloads for the attribute constructors that then immediately store the value into a `JsonNode` that's then passed into the intent.

JsonSchema/EvaluationResults.cs
```diff
@@ -421,13 +435,13 @@ public override void Write(Utf8JsonWriter writer, EvaluationResults value, JsonS
 		if (value.Format == OutputFormat.Hierarchical || value.Parent != null)
 		{
 			writer.WritePropertyName("keywordLocation");
-			options.Write(writer, value.EvaluationPath, JsonSchemaSerializerContext.Default.JsonPointer);
+			JsonSerializer.Serialize(writer, value.EvaluationPath, options);
```
jevansaks, line 438: Why did these switch back to JsonSerializer?

JsonSchema/JsonSchema.cs
```diff
@@ -814,7 +810,11 @@ public override void Write(Utf8JsonWriter writer, JsonSchema value, JsonSerializ
 		foreach (var keyword in value.Keywords!)
 		{
 			writer.WritePropertyName(keyword.Keyword());
-			options.Write(writer, keyword, keyword.GetType());
+
+			var keywordType = keyword.GetType();
+			var typeInfo = SchemaKeywordRegistry.GetTypeInfo(keywordType);
+			var converter = (IWeaklyTypedJsonConverter)typeInfo.Converter;
```
jevansaks, line 816: This is _slightly_ asymmetric with the read. options.Read handles the Converter not implementing IWeaklyTypedJsonConverter, but the Write doesn't. Might want to leverage the same kind of helper here.

JsonSchema/EvaluationResults.cs
gregsdennis: Not sure.  I'll put them back.

jevansaks (approved):
I left a couple comments that can be addressed now or later. Please update the PR description based on the work you've done too, it looks a little stale.

JsonSchema.Generation.Tests/TestEnvironment.cs
jevansaks: I saw that some of the object-based JsonNodes were coming through `public static JsonSchemaBuilder Const(this JsonSchemaBuilder builder, JsonNode? node)` too.

JsonSchema/EvaluationResults.cs
```diff
@@ -410,6 +407,8 @@ public override EvaluationResults Read(ref Utf8JsonReader reader, Type typeToCon
 	/// <param name="writer">The writer to write to.</param>
 	/// <param name="value">The value to convert to JSON.</param>
 	/// <param name="options">An object that specifies serialization options to use.</param>
+	[UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "We guarantee that the SerializerOptions covers all the types we need for AOT scenarios.")]
```
jevansaks, line 410: I suspect all the UnconditionalSuppressMessages can be removed

PR #649, opened 2024-02-02
Proposed fixes to get tests passing again

jevansaks:


gregsdennis:
I've got the fix for the remaining tests locally.  I'll pull this in, make the final fixes, and push that to #645.

Issue #652, opened 2024-02-09
Can we get JsonPath.Net Support for .NET Standard 2.0?


Emcrank:
### Documentation

- [X] I have consulted the [documentation](https://docs.json-everything.net), and my question isn't answered there.

### Nuget Package

JsonPath.Net

### Package Version

0.8.0

### How can I help?  Please provide as much context as possible.

Can we possibly get a package version of `JsonPath.Net` for `.NET Standard 2.0`? 

We are looking to get away from `Newtonsoft.Json` towards `System.Text.Json` however the latter does not have support for traversing the DOM using JsonPath's so I stumbled upon your library from the following [issue raised here](https://github.com/dotnet/runtime/issues/31068). Reason we cannot use it at .NET Standard 2.1 is we have some legacy application's which only accept imports for the latest .NET Framework. I've looked at the changes required and I don't think there is many, just mainly the fact there is no implicit operator to convert from string to `ReadOnlySpan<char>`. I suggest potentially calling `string.AsSpan()` in its place so it is compatible with .NET Standard 2.0?

Let me know your thoughts, cheers.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gregsdennis:
I tried to do that with the recent update but couldn't get it building.

I can try again next week.  Also happy to accept a PR.

Emcrank:
I've made the changes & all tests passing. Will submit a PR.

Emcrank:
https://github.com/gregsdennis/json-everything/pull/653

gregsdennis:
Merging the PR should close the issue. It should stay open until the problem is actually fixed.

gregsdennis:
Odd.  The PR merge should have closed this.

PR #653, opened 2024-02-10
Add support for .NET Standard 2.0 for JsonPath.Net

Emcrank:
Resolves https://github.com/gregsdennis/json-everything/issues/652

Adds target framework for netstandard2.0 for JsonPath.Net

### DIFF
--- a/JsonPath/Expressions/Operators.cs
+++ b/JsonPath/Expressions/Operators.cs
@@ -83,22 +83,22 @@ internal static class BinaryComparativeOperatorParser
 
 		var portion = source[index..(index + 2)];
 
-		if (portion.Equals("==", StringComparison.Ordinal))
+		if (portion.Equals("==".AsSpan(), StringComparison.Ordinal))
 		{
 			op = Operators.EqualTo;
 			index += 2;
 		}
-		else if (portion.Equals("!=", StringComparison.Ordinal))
+		else if (portion.Equals("!=".AsSpan(), StringComparison.Ordinal))
 		{
 			op = Operators.NotEqualTo;
 			index += 2;
 		}
-		else if (portion.Equals("<=", StringComparison.Ordinal))
+		else if (portion.Equals("<=".AsSpan(), StringComparison.Ordinal))
 		{
 			op = Operators.LessThanOrEqualTo;
 			index += 2;
 		}
-		else if (portion.Equals(">=", StringComparison.Ordinal))
+		else if (portion.Equals(">=".AsSpan(), StringComparison.Ordinal))
 		{
 			op = Operators.GreaterThanOrEqualTo;
 			index += 2;
@@ -113,7 +113,7 @@ internal static class BinaryComparativeOperatorParser
 			op = Operators.GreaterThan;
 			index++;
 		}
-		else if (portion.Equals("in", StringComparison.Ordinal))
+		else if (portion.Equals("in".AsSpan(), StringComparison.Ordinal))
 		{
 			op = Operators.In;
 			index += 2;
@@ -146,12 +146,12 @@ internal static class BinaryLogicalOperatorParser
 
 		var portion = source[index..(index + 2)];
 
-		if (portion.Equals("&&", StringComparison.Ordinal))
+		if (portion.Equals("&&".AsSpan(), StringComparison.Ordinal))
 		{
 			op = Operators.And;
 			index += 2;
 		}
-		else if (portion.Equals("||", StringComparison.Ordinal))
+		else if (portion.Equals("||".AsSpan(), StringComparison.Ordinal))
 		{
 			op = Operators.Or;
 			index += 2;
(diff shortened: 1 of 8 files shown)


JsonPath/Expressions/Operators.cs
```diff
@@ -83,22 +83,22 @@ public static bool TryParse(ReadOnlySpan<char> source, ref int index, [NotNullWh
 
 		var portion = source[index..(index + 2)];
 
-		if (portion.Equals("==", StringComparison.Ordinal))
+		if (portion.Equals("==".AsSpan(), StringComparison.Ordinal))
```
gregsdennis, line 86: Since these are constants, we should store the span representation in a field so it doesn't have to be recalculated repeatedly.

JsonPath/JsonPath.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFrameworks>netstandard2.1;net8.0</TargetFrameworks>
+		<TargetFrameworks>netstandard2.1;net8.0;netstandard2.0</TargetFrameworks>
```
gregsdennis, line 4: I'd prefer just replacing `netstandard2.1`.

JsonPath/PathSegment.cs
```diff
@@ -88,7 +88,8 @@ public override string ToString()
 		if (IsShorthand)
 			return $"{((IHaveShorthand)Selectors[0]).ToShorthandString()}";
 		
-		return $"[{string.Join(',', Selectors.Select(x => x.ToString()))}]";
+
```
gregsdennis, line 91: Please remove extra line.  (Single blank, not two.)

gregsdennis:
Thanks for this.  Tests look good.  Just a few things.

Emcrank:
Apologies, I was in a bit of a rush. Agree with all your comments. Will resolve these later tonight.

JsonPath/JsonPath.csproj
Emcrank: Tried removing netstandard2.1 target however this causes the build to fail... which I believe is due to the [NotNullWhen](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.notnullwhenattribute?view=netstandard-2.1) only being available in netstandard2.1 and beyond? Could be wrong though doesn't really make sense to me the error?

```JsonPath.Tests\Suite\CburgmerFeatureValidationTests.cs(139,54,139,65): error CS0433: The type 'NotNullWhenAttribute' exists in both 'JsonPath.Net, Version=0.8.0.0, Culture=neutral, PublicKeyToken=17ed63f672b0e278' and 'System.Runtime, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'```

gregsdennis: I wouldn't expect it to build against netstandard2.0, then.

Make sure that PolySharp is added to the project. There's a reference and a project property.  Check the other projects for examples.

Emcrank: The error relates to the test project, not the package library project.

I think it is getting confused with the multi targeting from a test perspective - It does build & tests all pass when all 3 target's are present. I presume when `netstandard2.1` is targeted, the tests are able to pull through the attribute because of that.. Not too sure if i'm honest.

PolySharp is added already to JsonPath project & csproj property is present.
`<PolySharpIncludeRuntimeSupportedAttributes>true</PolySharpIncludeRuntimeSupportedAttributes>`
`<PackageReference Include="PolySharp" Version="1.14.1" PrivateAssets="All" />`

JsonPath/PathSegment.cs
Emcrank: Resolved

JsonPath/Expressions/Operators.cs
Emcrank: Resolved

JsonPath/JsonPath.csproj
gregsdennis: @Sergio0694 do you know what's going on here?  I'd expect `[NotNullWhenTrue]` to be added for `netstandard2.0`.

gregsdennis: Oh!  It's because internals are visible to the tests project.

PolySharp is generating the attribute in _JsonPath.Net_ as an internal type in .Net Standard 2.0.  Then that version is being pulled for the .Net 6 version of the tests.  But since .Net 6 already defines the attribute, there's an ambiguity.

gregsdennis: @Emcrank just remove the attribute for now, please.  I'll figure it out later.

Emcrank: @gregsdennis Okay thanks, will do. Not familiar with the workings of PolySharp but makes sense.

Emcrank: That's committed.

gregsdennis:
Thanks for this @Emcrank

PR #660, opened 2024-02-18
Reusing schemas with different/changed options

gregsdennis:
@danielstarck I think this should resolve the issues that were highlighted by #608.


gregsdennis:
Also found an optimization around populating base URIs:

before:
image

after:
image

gregsdennis:
There may still be some optimizations available, but this fixes a bug, so I'm going to push it.

PR #670, opened 2024-02-23
add xml comment support

### LINKED ISSUE #667: Support xml comments on generated schema
### Nuget Package

JsonSchema.Net.Generation

### Package Version

4.0.0

I would be really nice if the doc comments on member/types could automatically be included when generating a json schema from a type.

I know i can use the `[description]` attribute, however as i already have xml comments it would be nice to not have to duplicate those :)

Using the `[description]` attribute. 

_No response_

### Code of Conduct

gregsdennis:
Resolves #667


JsonSchema.Generation.Tests/CommentsTests.cs
gregsdennis, line 1: @zlepper, please have a look at these tests to see if this aligns with your expected usage.

Basically, `Summary` becomes `description` unless a `[Description]` attribute is added, which will override the XML comments.

All of this only if a comments file has been registered in the config.

Optimization only pulls out references for "same" schemas, which basically means "type + member attributes".  So in the last test, a definition is created for `Inner2` and `Inner3`, but `Inner` doesn't reference the definition since it has an additional member attribute.  This is current behavior.

(I think properly, `Inner` should reference the definition with additional constraints, but there are some issues around arrays and attributes that are intended for the item type instead of the array itself.  It's a whole thing, and I don't feel like addressing it here.

gregsdennis:
I also need to figure out how to support this online.

JsonSchema.Generation.Tests/CommentsTests.cs
zlepper: From what i can tell, it looks exactly like what i would expect :D

PR #678, opened 2024-03-03
Add Italian localization

19bibo85:
Add resource and nuspec for Italian localization and update README.

### DIFF
--- a/JsonSchema/nuspec/JsonSchema.Net.it.nuspec
+++ b/JsonSchema/nuspec/JsonSchema.Net.it.nuspec
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package>
+  <metadata minClientVersion="2.12">
+    <id>JsonSchema.Net.it</id>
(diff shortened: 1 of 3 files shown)


JsonSchema/nuspec/JsonSchema.Net.it.nuspec
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package>
+  <metadata minClientVersion="2.12">
+    <id>JsonSchema.Net.it</id>
+    <version>2.0.0</version>
```
gregsdennis, line 5: Version should be 1.0.0 since this will be a new package.

gregsdennis (approved):
Thanks for this!

JsonSchema/nuspec/JsonSchema.Net.it.nuspec
19bibo85: You're right! I have updated the PR changing the version from `2.00` to `1.0.0`.

PR #706, opened 2024-04-11
Schema/lessons from functional

gregsdennis:
Importing some improvements found while attempting a new architecture (that didn't work out).

Primarily, schema registration now scans the schema and stores identifier information while also setting base URIs and spec versions.  With all of the identifer info stored in the schema registry, the reference keywords just need to go there.


gregsdennis:
I'm not sure what's going on here, but performance isn't improving, and I still need to jump through hoops to get everything initialized right.  Closing for now, but I will still continue looking at performance improvement opportunities.

PR #707, opened 2024-04-16
IReadOnlyCollections to IReadOnlyLists in KeywordAccess extensions

SGStino:
Keywords that are list based might benefit for their extensions to also be list based.


gregsdennis:
This looks good.  Thanks.  I'll pull it into a new branch to update versions, etc.

PR #712, opened 2024-04-18
Nuget updates

### LINKED ISSUE #711: Update Nuget package meta-data
### Summary
I wish for the nuget packages to have the licence expression property set correctly

### Details

The licence expression property should be set to the correct licence type I.e. MIT as this will enable analysis of licences in use to occur in external tools & the license type will be shown in Nuget etc.

gregsdennis:
Resolves #711

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,7 @@ indent_style = tab
 indent_size = tab
 tab_width = 4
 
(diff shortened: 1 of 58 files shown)


gregsdennis:
Alright... let's hope this works!

PR #714, opened 2024-04-19
Do not evaluate untaken if and switch branches

### LINKED ISSUE #713: Error "unknown context value" Raised After Checking for Existence
### Nuget Package

JsonE.Net

### Package Version

2.1.0

### Operating System

Windows

.Net (5 or after)

After an if (or switch) statement that checks if a variable/key exists, the `then` statement throws an `unknown context value` error even though the `if` check was false. For example, the [sample for `defined()`](https://json-e.js.org/built-ins.html#context) on the json-e page works in the online playground there but causes this error in JsonE.Net.

### What did you expect?

When an `if` or `switch` statement evaluates false, errors should not be thrown for non-existent values in the untaken branch.

Here are a few tests in the `more-tests.yml` format that fail, and my understanding from reading the spec and using the online [playground](https://json-e.js.org/playground.html) is that they should pass:

```yaml
---
title:    if defined check
context:  {y: 10}
template: {$if: 'defined("x")', then: {$eval: 'x'}, else: 20}
result:   20
---
title:    switch defined check
context:  {y: 10}
template: {$switch: {'defined("x")': {$eval: 'x'}, $default: 20}}
result:   20
```

The `template` can be copied to the playground to verify that at least in *their* implementation, this is supported.

If this issue is accepted, it'd be great for these tests to be incorporated in the official json-e test suite in case it exists in other implementations.

mi25iw:
Resolves #713

The samples for json-e indicate that one can check for an value in an if statement and attempt to use it in the 'then' block.

This change delays evaluating the 'then' and 'else' blocks until they are needed, and a similar change is applied to the 'switch' statement.

The change for `IfThenElseOperator` is fairly straightforward. The `SwitchOperator` one was as well, although it does change the order from:

1. Evaluate parameter dictionary (`JsonE.Evaluate(parameter, context)!.AsObject();`)
2. Check truthiness
3. Copy value

to:

1. Iterate over parameter dictionary (keys have not yet went to `JsonE.Evaluate`! Not sure if this is important. I saw one code path that looked like it might unescape keys or impact it another way. Please review carefully.)
2. Check truthiness
3. Evaluate value (and still copy it since the original code did, I'm new to the codebase so that might not be necessary).

It passed all the tests still, but I am not super comfortable that I changed the order above. I'm hoping you know (or have any easy way to check) if that's problematic.

### DIFF
--- a/JsonE.Tests/Files/more-tests.yml
+++ b/JsonE.Tests/Files/more-tests.yml
@@ -146,3 +146,13 @@ title:    accessor after function call
 context:  {}
 template: {$eval: 'split("left:right", ":")[1]'}
 result:   right
+---
+title:    if defined check
+context:  {y: 10}
+template: {$if: 'defined("x")', then: {$eval: 'x'}, else: 20}
+result:   20
+---
+title:    switch defined check
+context:  {y: 10}
+template: {$switch: {'defined("x")': {$eval: 'x'}, $default: 20}}
+result:   20
--- a/JsonE/Operators/IfThenElseOperator.cs
+++ b/JsonE/Operators/IfThenElseOperator.cs
@@ -26,11 +26,8 @@ internal class IfThenElseOperator : IOperator
 		var thenPresent = obj.TryGetValue("then", out var thenValue, out _);
 		var elsePresent = obj.TryGetValue("else", out var elseValue, out _);
 
-		thenValue = JsonE.Evaluate(thenValue, context);
-		elseValue = JsonE.Evaluate(elseValue, context);
-
 		return cond.IsTruthy()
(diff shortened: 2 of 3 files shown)


gregsdennis:
Nice! I'll pull this into an update branch to do version and release note stuff.  It'll publish when that merges.

gregsdennis:
The copying is due to `JsonNode` only being allowed to have a single parent.  It's generally necessary, but I don't like it.  I could do some inspection to determine whether the specific situation calls for it, but... meh.

mi25iw:
Makes sense. Thanks for the quick response (and for the libraries in general)!

PR #717, opened 2024-04-20
fix an exception message

gregsdennis:
Missed this type rename during the AOT update.


Andree778 (commented):
[rn-json-path.md](https://github.com/user-attachments/files/17076545/rn-json-path.md)

PR #729, opened 2024-05-13
chore: Update sliceselector to match the rfc

He-Pin:
Motivation:
Update the slice selector to match the rfc

https://datatracker.ietf.org/doc/html/rfc9535#section-2.3.4.2.2

### DIFF
--- a/JsonPath/SliceSelector.cs
+++ b/JsonPath/SliceSelector.cs
@@ -57,7 +57,7 @@ public class SliceSelector : ISelector
 		if (Step == 0) yield break;
 
 		var step = Step ?? 1;
-		var start = Start ?? (step >= 0 ? 0 : arr.Count);
+		var start = Start ?? (step >= 0 ? 0 : arr.Count - 1);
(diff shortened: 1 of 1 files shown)


gregsdennis:
@He-Pin Have you run the tests?

Also, please open an issue to discuss changes before submitting a PR.  It saves you work in the end.

gregsdennis:
The current tests pass, but I'd be interested whether you can create a test that doesn't pass that your change fixes.  Otherwise, I'm inclined to leave it as is.

He-Pin:
image

@gregsdennis I will do after work with a local IDE:)

gregsdennis:
That's fine, but it works as is, and it was finicky to get working for some reason.  So unless there's a case for which the current code fails, I'm inclined to not mess with it.

gregsdennis:
@He-Pin are you going to continue with this?  Otherwise, I'm going to close it.

He-Pin:
I can try it this weekend

PR #732, opened 2024-05-14
Restructure repo

gregsdennis:
Consolidating library projects and their tests into a `src/` folder in order to clean up the repo landing page and make the readme more accessible.


gregsdennis:
Not sure why the labeler isn't picking up the changes in the project folders.  Will debug that when I update a package more directly.

PR #734, opened 2024-05-15
Website updates

gregsdennis:
Does two things primarily:

- Add a sponsor cloud (for when I get sponsors at sufficient tiers)
- Updates landing page to list features (to make sponsor cloud less lonely)


gregsdennis:
Well... let's hope.

PR #738, opened 2024-05-17
Address AOT Trim warning in JsonPath

jevansaks:
### Description

Fix AOT trim warning from JsonPath (`'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods' in call to 'System.Type.GetMethods(BindingFlags)'. The return value of method 'System.Object.GetType()' does not have matching annotations.`).

* Added annotations to FindEvaluationMethods (internal method)
* Added new public APIs with overloads that have the appropriate AOT annotations.

### Links

I didn't file a separate issue to report this.

### Checks

- [x] I have read and understand the [Code of Conduct](https://github.com/gregsdennis/json-everything/blob/master/CODE_OF_CONDUCT.md) and [Contribution Guidelines](https://github.com/gregsdennis/json-everything/blob/master/CONTRIBUTING.md).

### Details:

I was trying out JsonPath in an AOT project and hit a trim warning during publish:

```
src\JsonPath\FunctionRepository.cs(94,17,94,85): warning IL2075: 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods' in call to 'System.Type.GetMethods(BindingFlags)'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
```

That was caught by the NAOT trimmer that wasn't seen by the static analyzer for .NET 8. Luckily, this static analysis bug is fixed in .NET 9, so added "net9" as a TargetFramework in the csproj and the warning shows up.

I rewrote the code slightly to carry the annotation along per https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/fixing-warnings#dynamicallyaccessedmembers.

### DIFF
--- a/src/JsonPath/FunctionRepository.cs
+++ b/src/JsonPath/FunctionRepository.cs
@@ -16,44 +16,107 @@ public static class FunctionRepository
 
 	static FunctionRepository()
 	{
-		Register(new LengthFunction());
-		Register(new CountFunction());
-		Register(new MatchFunction());
-		Register(new SearchFunction());
-		Register(new ValueFunction());
+		RegisterValueFunction<LengthFunction>();
+		RegisterValueFunction<CountFunction>();
+		RegisterLogicalFunction<MatchFunction>();
+		RegisterLogicalFunction<SearchFunction>();
+		RegisterValueFunction<ValueFunction>();
 	}
 
 	/// <summary>
 	/// Registers a new function implementation, allowing it to be parsed.
 	/// </summary>
 	/// <param name="function">The function.</param>
+	[RequiresUnreferencedCode("Use the overloads that take a Type for AOT compatibility")]
 	public static void Register(ValueFunctionDefinition function)
 	{
-		_functions[function.Name] = function;
-
-		FindEvaluationMethods(function, typeof(JsonNode));
+		Register(function, function.GetType());
 	}
 
 	/// <summary>
 	/// Registers a new function implementation, allowing it to be parsed.
 	/// </summary>
 	/// <param name="function">The function.</param>
+	/// <param name="functionType">The type of the function.</param>
+	public static void Register(ValueFunctionDefinition function,
+		[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)] Type functionType)
+	{
+		_functions[function.Name] = function;
+
+		FindEvaluationMethods(function, functionType, typeof(JsonNode));
+	}
+
+	/// <summary>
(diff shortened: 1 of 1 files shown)


jevansaks:
Sorry, this was incomplete. Flipping to draft while I sort it out.

gregsdennis:
Thanks for this. I'll look at it when I can.

Is there a way to test this besides adding a new framework? 

Also, I recently added a PR template. Did it not come up for you?

jevansaks:
> Is there a way to test this besides adding a new framework?

It shows up if you make an app that's `<PublishAot>` and then you publish. We had been avoiding doing that since the goal with .NET's static analysis is that all the warnings _should_ show up when compiling the library as IsAotCompatible. And the fact that this warning didn't show up was just a bug, so I'm inclined to leave it and just add net9 as a target framework when it's stable.

> Also, I recently added a PR template. Did it not come up for you?

Oops, I deleted too much of it. Fixed up the PR description.

src/JsonPath/FunctionRepository.cs
```diff
@@ -16,44 +16,107 @@ public static class FunctionRepository
...
+		Register(function, function.GetType());
+	}
+
+	/// <summary>
+	/// Registers a new function implementation, allowing it to be parsed.
+	/// </summary>
+	/// <param name="function">The function.</param>
+	/// <param name="functionType">The type of the function.</param>
+	public static void Register(ValueFunctionDefinition function,
+		[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)] Type functionType)
```
gregsdennis, line 42: I don't understand how this fixes the problem.  Isn't this marking the method (via the param) as non-AOT?

jevansaks: Reflection is compatible with AOT (all metadata for all types is included by default), but not compatible with trimming. Trimming is another layer on top of AOT where you can reduce the size of the final binary by trimming code/data out that the compiler can tell is unreachable. For reflection, metadata is trimmed out by default _unless_ the compiler can see statically that that some code needs the metadata.

These attributes (as described [here](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming)) are how you annotate `Type` parameters or template arguments with which aspects of metadata the function is going to need.

gregsdennis:
Thanks.  I'll update version and release notes on a new branch, then push it out.

jevansaks:
Thanks!

PR #741, opened 2024-05-25
chore: Fix typo in InOperator

He-Pin:
### Description
fix typo, it should be `in`

### DIFF
--- a/src/JsonPath/Expressions/InOperator.cs
+++ b/src/JsonPath/Expressions/InOperator.cs
@@ -31,6 +31,6 @@ internal class InOperator : IBinaryComparativeOperator
 
 	public override string ToString()
 	{
-		return "in";
+		return " in ";
 	}
 }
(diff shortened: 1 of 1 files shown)


gregsdennis:
Nice catch.  I'll roll this into the next version.

src/JsonPath/Expressions/InOperator.cs
```diff
@@ -31,6 +31,6 @@ public bool Evaluate(PathValue? left, PathValue? right)
 
 	public override string ToString()
 	{
-		return "<";
+		return "in";
```
gregsdennis, line 34: Can you add a test in a new test class called `StringifyTests`, please?  I'll fill in others that are missing afterward.

gregsdennis (approved):

src/JsonPath/Expressions/InOperator.cs
gregsdennis: I think this may need surrounding spaces:

```suggestion
		return " in ";
```

To be confirmed with a test, but I would expect the syntax to require them.

He-Pin: @gregsdennis I think I may not be able to ,because I'm a daily Scala/Java user, sorry for that.

PR #751, opened 2024-06-23
improve term grouping logic for filter expressions

gregsdennis:
Improves path stringification so that the stringified path can be parsed back to a path that returns the same results as the original.  This allows for variances like whitespace.


gregsdennis:
Looks like I need to refactor the parsing logic to handle some of this.

gregsdennis:
For reference, from previous PR:

image

PR #773, opened 2024-07-25
Improve performance of regex usage in JsonSchema

### LINKED ISSUE #761: Pattern and PatternProperties keyword performance and security improvements
### Nuget Package

JsonSchema.Net

### Package Version

7.1.4

The PatternKeyword and the PatternPropertiesKeyword implementations both create new compiled `Regex`  instances.

According to best practices as described here: https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-regex#static-regular-expressions regex instances that are often used should be cached for performance benefits.

To achieve this, the keywords could use the static `Regex.IsMatch` method to use the internal cache (optionally add an entry in the `readme`  file to increase the [Regex.CacheSize](https://learn.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.cachesize?view=net-8.0) property.

Another options is to implement a regex cache within the JsonSchema.net library.

Also, `Regex` instances [should set a timeout](https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-regex#use-time-out-values) especially because json schema's will validate untrusted user input.

I'd be happy to create a PR for this, but that would require breaking changes to the `PatternKeyword` and the `PatternPropertiesKeyword` I think. 

_No response_

_No response_

### Code of Conduct

rikbosch:
### Description

Improved usage of `Regex` in several places.

When a `string` pattern is supplied for `Pattern` or `PatterProperties` keywords, the static `Regex.IsMatch` will be used to evaluate the keywords. 

When a `Regex` instance is supplied, that instance is used by the keywords.

I also made the `_dateTimeRegex` a compiled regex to improve performance. (this could be improved further by using code generated regex instances.)

Updated the `Formats` class to use code generated regex if possible.

## breaking changes

* The return type of `JsonSchemaExtensions.KeywordAccess` `GetPattern` and `GetPatternProperties` will return the `string` pattern and not the `Regex` instance.
* The  `Regex Value` property in the `PatternKeyword` was removed, in favor of a `string Pattern` property. 
* The `PatternPropertiesKeyword.Patterns` property changed from an `IReadOnlyDictionary<Regex, JsonSchema>` to ` IReadOnlyDictionary<string, JsonSchema>`

### Links

Resolves #761  

### Checks

### DIFF
--- a/src/JsonSchema.DataGeneration/Generators/StringGenerator.cs
+++ b/src/JsonSchema.DataGeneration/Generators/StringGenerator.cs
@@ -66,7 +66,7 @@ internal class StringGenerator : IDataGenerator
 			//if (context.AntiPatterns != null)
 			//	overallRegex += HelperExtensions.Forbid(context.AntiPatterns.Select(x => x.ToString()));
 			if (context.Pattern != null)
-				overallRegex = context.Pattern.ToString();
+				overallRegex = context.Pattern;
 
 			Console.WriteLine(overallRegex);
 
--- a/src/JsonSchema.DataGeneration/Requirements/StringRequirementsGatherer.cs
+++ b/src/JsonSchema.DataGeneration/Requirements/StringRequirementsGatherer.cs
@@ -32,7 +32,7 @@ internal class StringRequirementsGatherer : IRequirementsGatherer
 			supportsStrings = true;
 		}
 
-		var pattern = schema.Keywords?.OfType<PatternKeyword>().FirstOrDefault()?.Value;
+		var pattern = schema.Keywords?.OfType<PatternKeyword>().FirstOrDefault()?.Pattern;
 		if (pattern != null)
 		{
 			//context.Patterns ??= new List<Regex>();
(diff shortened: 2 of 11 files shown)


gregsdennis:
Thanks for this.

I think some simple changes could avoid the breaks.

- In `JsonSchemaExtensions.KeywordAccess`, `.GetPattern()` and `.GetPatternProperties()` could be `.GetPatternValue()` and `.GetPatternPropertiesValues()`, respectively, and we keep the existing methods.
- Keep `PatternKeyword.Value`.
- Keep `PatternPropertiesKeyword.Patterns` as it is and create a new `PatternPropertiesKeyword.PatternValues`.

Then mark all of the existing properties as obsolete and set them up to lazy-initialize.  (Please use `_field ??= GetField()` instead of `Lazy<T>`.)

Some of this may not "feel" quite right, but I'd prefer to avoid the breaking change until I can't.

gregsdennis:
@rikbosch Please see my comments above.  I think we can do this without breaking changes.

rikbosch:
@gregsdennis done! And I agree with the breaking changes. I like your suggestion about the naming of the new properties / methods.

gregsdennis:
Thanks for this @rikbosch!  I'm going to pull this into another branch to do some personal preference cleanup, then I'll merge to master.  It'll push out an update at that point.

gregsdennis:
@rikbosch I have a question on the other PR ☝️.

PR #776, opened 2024-08-05
Schema/improved regex handling

### LINKED ISSUE #761: Pattern and PatternProperties keyword performance and security improvements

gregsdennis:
### Description

Updating details from #773 

### Links

Resolves #761

### Checks

### DIFF
--- a/src/JsonSchema/PatternKeyword.cs
+++ b/src/JsonSchema/PatternKeyword.cs
@@ -35,6 +35,12 @@ public class PatternKeyword : IJsonSchemaKeyword
 
 	public string Pattern => _regexOrPattern;
 
+	/// <summary>
+	/// Returns the Regex Value of the keyword
+	/// </summary>
+	[Obsolete($"Please use the '{nameof(Pattern)}' property instead.")]
+	public Regex Value => _regexOrPattern.ToRegex();
+
 
 	/// <summary>
 	/// Creates a new <see cref="PatternKeyword"/> based on a regular expression pattern.
--- a/src/JsonSchema/PatternKeyword.cs
+++ b/src/JsonSchema/PatternKeyword.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Text.Json;
 using System.Text.Json.Serialization;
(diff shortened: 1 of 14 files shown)


src/JsonSchema/PatternPropertiesKeyword.cs
```diff
@@ -25,43 +25,65 @@ namespace Json.Schema;
...
+	public IReadOnlyDictionary<string, JsonSchema> PatternValues => _patternValues ??= _patternsLookup.ToDictionary(x => x.Key, x => x.Value.Schema);
+	
 	/// <summary>
-	/// If any pattern is invalid or unsupported by <see cref="Regex"/>, it will appear here.
+	/// The regex patterns of this PatternPropertiesKeyword
 	/// </summary>
-	/// <remarks>
-	/// All validations will fail if this is populated.
-	/// </remarks>
-	public IReadOnlyList<string>? InvalidPatterns { get; }
```
gregsdennis, line 46: @rikbosch how is this being handled now?

src/JsonSchema/PatternKeyword.cs
```diff
@@ -23,39 +24,44 @@ namespace Json.Schema;
...
-	/// </summary>
-	public Regex Value { get; }
-	/// <summary>
-	/// If the pattern is invalid or unsupported by <see cref="Regex"/>, it will appear here.
+	/// Returns the Regex Value of the keyword
 	/// </summary>
-	/// <remarks>
-	/// All validations will fail if this is populated.
-	/// </remarks>
-	public string? InvalidPattern { get; }
```
gregsdennis, line 41: Same here.

gregsdennis: I've added a couple tests to `master` and rebased to show this.

gregsdennis: If `RegexOrPattern` were to check the pattern in the constructor, would that insert the pattern into the Regex cache (what this whole change was trying to do anyway)?

> I would say that it's an illegal json schema when the regex pattern is invalid, so an exception when deserializing json to a JsonSchema sounds valid.

I want to say there were some optional tests that had valid regexes that .Net just didn't support.  It seems the only one that fits that now are the UTF-16 tests, but even those just don't evaluate correctly.

Moreover, the `.Pattern()` builder extension _does_ throw an exception on a bad pattern as well.

I think I want to change the behavior long term, but I'd still like to leave the current support for now.  I'll work something out.

gregsdennis: I suppose I can include this change and say in the release notes that it's fixing a bug.  Doesn't really explain why the `InvalidPattern` property was there in the first place, though.

rikbosch: Yeah, performing an `IsMatch` in the constructor works as well, and is actually closer to the original behavior.

PR #781, opened 2024-08-14
Count subschemas before memory allocation

### LINKED ISSUE #778: IndexOutOfRangeException when evaluating against schema with more than 512 definitions
### Nuget Package

JsonSchema.Net

### Package Version

7.x

### Operating System

None

None

I have a very big schema file which consists of many schemas ($defs)
When trying to evaluate against this schema, I get an `IndexOutOfRangeException`. The exception occurs in [JsonSchema.cs line 571](https://github.com/json-everything/json-everything/blob/b5e3964ffad512cc3531cddd65dd68804ae85856/src/JsonSchema/JsonSchema.cs#L571) because the span used there has a default size of 512, which is not enough in my case.

Validating with this schema does work in 5.x and 6.x

### What did you expect?

I did expect the schema to successfully validate.

Here is a dummy schema, that will cause the exception. The schema itself is kind of useless, I just put it together to demonstrate the issue.
[big-schema.json](https://github.com/user-attachments/files/16510360/big-schema.json)

``` CSHARP
var schema = JsonSchema.FromFile("C:\\path\\to\\big-schema.json");
var stream = new FileStream("C:\\path\\to\\any.json", FileMode.Open);
var document = await JsonDocument.ParseAsync(stream); // <-- exception occurs here
```

Perhaps the buffer size used here could be configured and when the buffer is not big enough an exception that explains the problem is thrown.
The default buffer of 512 probably is sufficient in most cases.

Perhaps a simple static configuration where the application can just call something like
```CSHARP
JsonSchemaConfiguration.MinimumSchemaBufferSize = 4096;
```
on startup.

gregsdennis:
### Description

Schemas which have more than 512 subschemas will throw an `IndexOutOfRangeException` during registration.

cc: @Laniusexcubitor

### Links

Resolves #778     

### Checks

### DIFF
--- a/src/JsonSchema/EvaluationContext.cs
+++ b/src/JsonSchema/EvaluationContext.cs
@@ -50,15 +50,6 @@ public class EvaluationContext
 		Scope = new DynamicScope(initialScope);
 	}
 
-	internal static IEnumerable<IJsonSchemaKeyword> GetKeywordsToProcess(JsonSchema schema, EvaluationOptions options)
-	{
-		if (options.ProcessCustomKeywords ||
-		    schema.Dialect == null) return schema.Keywords!;
-
-		var vocabKeywordTypes = schema.Dialect.SelectMany(x => x?.Keywords ?? []);
-		return schema.Keywords!.Where(x => vocabKeywordTypes.Contains(x.GetType()));
-	}
-
 	internal void PushEvaluatingAs(SpecVersion version)
 	{
 		_evaluatingAs.Push(version);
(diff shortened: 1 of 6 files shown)


gregsdennis:
I'm not sure what the failure is here.  It seems that when I run the entire test suite, I get a failure, but any portion of it (including various combinations of subsuites) passes.  I'm not sure how the latest commit could affect it.  Maybe it's just flaky.

gregsdennis:
Looking at the test run from the Action, it looks like it's just that one flaky datagen test.

PR #802, opened 2024-10-01
add action to test signing a package

gregsdennis:
Just testing package signing.  Move along.


glennawatson:
just got back from holidays, will take a look tomorrow for you.

PR #822, opened 2024-11-30
Version updates

### LINKED ISSUE #821: How to make NodeListFunctionDefinition work?
### Documentation

- [X] I have consulted the [documentation](https://docs.json-everything.net), and my question isn't answered there.

### Nuget Package

JsonPath.Net

### Package Version

1.1.6

### How can I help?  Please provide as much context as possible.

I'm trying to extend the existing functions with a function of type NodeType, but cannot figure out how to make it work. I'm unsure if it's a bug or has not yet been implemented, as it seems that only the logical functions are referring back to the FunctionRepository, and as the NodeListFunctionDefinition isn't a logical one, that fails to parse the path.

Here's the definition I'm trying to use (minimalistic implementation):

```
 public class MagicFunction : NodelistFunctionDefinition
    {
        public override string Name => "magic";

        public NodeList? Evaluate(NodeList? nodeList)
        {
            return nodeList;
        }
    }
```

I registered it via `FunctionRepository.RegisterNodelistFunction<MagicFunction>();` which seems to work fine.

Then I try to parse a JsonPath, which fails. I haven't found other examples for NodeType functions anywhere else, so maybe the path is wrong?

```
const string jsonPath = "$.foo[?magic(@)].bar";
var path = JsonPath.Parse(jsonPath); // throws an Exception
```

I'm trying something pretty special and a valid answer to my question might also be that I'm going at it all wrong :) So here's what I'm trying to do: For the path above, I would like my new nodelist function to take the matches up to that part (so `.foo`) and then return a fresh JSON (from outside) that will include the `.bar`. 

### Code of Conduct

gregsdennis:
### Description

- Adds .Net 9 support to all libs.
- Updates _System.Text.Json_ to v9.
- Updates all test projects from .Net 6/8 to .Net 8/9.
  - .Net 6 was being used to test the .Net Standard 2.0 version of the libs, but that has dropped out of support now.  I tried adding .Net Framework, but there were several issues (example), and I decided to simply leave it untested.  The libs continue to publish as .Net Standard to provide some support for .Net Framework, but I provide no guarantees.  I figure that usage will be edge case anyway.
- Fixes #821 by adding support for (and testing) nodelist functions as an existence check.
- Remove obsolete JsonPath.PathResult.Errors: Evaluation of a valid path never returns errors (per spec)
  - Major bump for JsonPath.Net and JsonSchema.Data

### Links

Resolves #821

### Checks


gregsdennis:
I think just all the release notes, then this can go.

PR #830, opened 2024-12-07
chore: Remove unused ExistsOperator.cs and NoOpOperator.cs

He-Pin:
Motivation:
These classes are never used.


He-Pin:
@dotnet-policy-service agree

He-Pin:
@dotnet-policy-service agree

PR #831, opened 2024-12-07
fix: FunctionValueExpressionParser should support NodeListFunction

He-Pin:
Motivation:
Make FunctionValueExpressionParser support NodelistFunctionDefinition too

### DIFF
--- a/src/JsonPath/Expressions/FunctionValueExpressionNode.cs
+++ b/src/JsonPath/Expressions/FunctionValueExpressionNode.cs
@@ -74,14 +74,14 @@ internal class FunctionValueExpressionParser : IValueExpressionParser
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


gregsdennis:
No, a nodelist is not a value.  This parser handles only functions which return values.

He-Pin:
But it's been evaluated

gregsdennis:
Add a test that fails without your change, please.  I don't think this is necessary.

He-Pin:
Let me check the `noop` first.

He-Pin:
I try to run parse `$[?count(noop(@.*))>0)]` but it fails,

gregsdennis:
Okay.  Good.  Please create a test, show that it fails without your changes and then apply your changes to show that they fix the problem.

gregsdennis:
Closing in favor of #835

PR #835, opened 2024-12-15
chore: Test noop nodelist method

He-Pin:
Motivation:
Show why https://github.com/json-everything/json-everything/pull/831 is needed

### DIFF
--- a/src/JsonPath.Tests/ExpressionTests.cs
+++ b/src/JsonPath.Tests/ExpressionTests.cs
@@ -53,7 +53,7 @@ public class ExpressionTests
 		FunctionRepository.RegisterNodelistFunction<NoOpFunction>();
 
 		// should do the same thing as $[?@.*]
-		var path = JsonPath.Parse("$[?noop(@.*)]");
+		var path = JsonPath.Parse("$[?noop(noop(@.*))]");
 		var data = JsonNode.Parse(
 			"""
 			[


He-Pin:
@gregsdennis ping~

src/JsonPath.Tests/ExpressionTests.cs
```diff
@@ -53,7 +53,7 @@ public void ExpressionWithNoOpFunctionWorks()
 		FunctionRepository.RegisterNodelistFunction<NoOpFunction>();
 
 		// should do the same thing as $[?@.*]
-		var path = JsonPath.Parse("$[?noop(@.*)]");
```
gregsdennis, line 56: Don't change existing tests.  Create new tests.

gregsdennis:
This is your new PR.

Steps are:

1. Create an issue so changes can be discussed (you didn't do this)
2. Once discussed, create a PR with a test that illustrates the problem.  The build should fail for this.
3. Apply one or more commits that fix the problem.  The build passes.

src/JsonPath.Tests/ExpressionTests.cs
He-Pin: Have some network issue (GFW), will update later.

gregsdennis:
I don't need to be pinged.

PR #843, opened 2024-12-27
Use 'params ReadOnlySpan' to reduce allocations on .NET9

cptjazz:
Use 'params ReadOnlySpan' in three situations, twice when creating JsonPointer from segments and once when parsing a JsonPath, to avoid array allocations on .NET9.

### DIFF
--- a/src/JsonPointer/JsonPointer.cs
+++ b/src/JsonPointer/JsonPointer.cs
@@ -205,6 +205,9 @@ public class JsonPointer : IEquatable<JsonPointer>, IReadOnlyList<string>
 	/// <remarks>This method creates un-encoded pointers only.</remarks>
 	public static JsonPointer Create(params PointerSegment[] segments)
 	{
+#if NET9_0_OR_GREATER
+		return Create(segments.AsSpan());
+#else
 		var array = new string[segments.Length];
 
 		for (var i = 0; i < segments.Length; i++)
@@ -213,7 +216,30 @@ public class JsonPointer : IEquatable<JsonPointer>, IReadOnlyList<string>
 		}
 
 		return new JsonPointer(array);
+#endif
 	}
+
+#if NET9_0_OR_GREATER
+
+	/// <summary>
+	/// Creates a new JSON Pointer from a collection of segments.
+	/// </summary>
+	/// <param name="segments">A collection of segments.</param>
+	/// <returns>The JSON Pointer.</returns>
+	/// <remarks>This method creates un-encoded pointers only.</remarks>
+	public static JsonPointer Create(params ReadOnlySpan<PointerSegment> segments)
+	{
+		var array = new string[segments.Length];
+		
+		for (var i = 0; i < segments.Length; i++)
+		{
+			array[i] = segments[i].Value;
+		}
+
+		return new JsonPointer(array);
+	}
+
+#endif
 
 	/// <summary>
 	/// Generates a JSON Pointer from a lambda expression.
@@ -295,7 +321,11 @@ public class JsonPointer : IEquatable<JsonPointer>, IReadOnlyList<string>
 
 		i++;
 
+#if NET9_0_OR_GREATER
+		return Create(segments[i..]);
+#else
 		return Create(segments[i..].ToArray());
+#endif
 	}
 
 	/// <summary>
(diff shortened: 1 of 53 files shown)


cptjazz:
| Method                | Count | Mean       | Error     | StdDev    | Gen0   | Allocated |
|---------------------- |------ |-----------:|----------:|----------:|-------:|----------:|
| CreateFromSegments    | 10    |   555.8 ns |  10.27 ns |  11.83 ns | 0.6495 |    1360 B |
| CreateFromSegmentsNew | 10    |   370.2 ns |   7.31 ns |  10.25 ns | 0.4206 |     880 B |
| CreateFromSegments    | 100   | 5,323.0 ns | 103.10 ns | 157.45 ns | 6.5002 |   13600 B |
| CreateFromSegmentsNew | 100   | 3,538.1 ns |  68.12 ns |  72.88 ns | 4.2076 |    8800 B |

for a benchmark of the form

````
{
	JsonPointer p = null;
	for (int i = 0; i < Count; i++)
		p = JsonPointer.Create("a", "b", "c");
	
	return p;
}
````

File: src/JsonPointer/JsonPointer.cs [hunk 1]
```diff
@@ -205,6 +205,9 @@ public static bool TryParse(string source, [NotNullWhen(true)] out JsonPointer?
 	/// <remarks>This method creates un-encoded pointers only.</remarks>
 	public static JsonPointer Create(params PointerSegment[] segments)
 	{
+#if NET9_0_OR_GREATER
+		return Create(segments.AsSpan());
```
gregsdennis, line 209: This doesn't make sense to me.  You're not saving anything because you've already allocated the array that was passed in.  You don't gain anything by calling the span version here.

src/JsonPointer/JsonPointer.cs
```diff
@@ -322,6 +352,9 @@ public JsonPointer Combine(JsonPointer other)
 	/// <returns>A new pointer.</returns>
 	public JsonPointer Combine(params PointerSegment[] additionalSegments)
 	{
+#if NET9_0_OR_GREATER
+		return Combine(additionalSegments.AsSpan());
```
gregsdennis, line 356: Same here.  The array is already allocated.  Calling the span version doesn't give you anything.

gregsdennis (commented):
In the future, please first open an issue so that changes can be discussed.  (It's in the contribution guidelines.)

Also, please use the PR template provided instead of deleting it.  It's there for a reason.

cptjazz:
Sorry about the formalities, I will stick to the guidelines next time

src/JsonPointer/JsonPointer.cs
cptjazz: You are right, it doesn't give me any benefit in terms of runtime/allocations, but instead of having 2 versions of basically the same method I can save in terms of code size, JIT time and code maintenance. That's the main reason I did it, since the AsSpan() is virtually free here.

src/JsonPointer/JsonPointer.cs
cptjazz: See above

gregsdennis:
I like the changes here, and I'd like to see them applied elsewhere if possible.  Have you checked in JsonSchema.Net for other opportunities?

cptjazz:
I was primarily looking at the parts that directly influence JsonSchema, since speeding up schema validation is my main goal. There is usage of `params` in JsonSchema but nothing that makes a noticeable difference outside of test code, as far as I can tell at the moment.
But there are opportunities in JsonE, JsonPatch, and JsonLogic.

I will add a few commits to target those.

cptjazz:
I added a few changes. There are more opportunities, but there we will have to discuss introducing breaking changes, since the public surface is quite big.

gregsdennis:
Thanks for this.  I'll pull it into #848 so I can update the release notes.

gregsdennis:
Interesting: R# doesn't like it.

image

cptjazz:
Do you have the newest R#? Allegedly it's fixed already: https://youtrack.jetbrains.com/issue/RSRP-499333/New-params-collections-in-.NET-9-are-marked-as-error-in-Extension-Methods

gregsdennis:
I hadn't updated, no.  Just thought it was funny.  I saw that update icon a few days ago and subsequently ignored it.

PR #850, opened 2025-01-03
Properly dispose of rented memory in JsonPointer.ToString()

### LINKED ISSUE #849: JsonPointer.ToString() -- rented memory is not disposed of
### Nuget Package

JsonPointer.Net

### Package Version

5.2.0

### Operating System

None

None

In the loop in JsonPointer.ToString() memory is rented, but not properly disposed of:

    var localOwner = MemoryPool<char>.Shared.Rent();

where it should be:

    using var localOwner = MemoryPool<char>.Shared.Rent();

### What did you expect?

Proper disposal of rented memory.

--

_No response_

cptjazz:
### Description
Properly disposes of the rented memory, and also sets the minimum buffer size when renting, to avoid (some) buffer resizings, since we already have the segment length at hand.

### Links

Fixes #849

### Checks

### DIFF
--- a/src/JsonPointer/JsonPointer.cs
+++ b/src/JsonPointer/JsonPointer.cs
@@ -10,7 +10,6 @@ using System.Reflection;
 using System.Text.Json;
 using System.Text.Json.Nodes;
 using System.Text.Json.Serialization;
-using System.Web;
 using Json.More;
 
 namespace Json.Pointer;
@@ -28,7 +27,6 @@ public class JsonPointer : IEquatable<JsonPointer>, IReadOnlyList<string>
 	public static readonly JsonPointer Empty = new();
 
 	private readonly string[] _decodedSegments;
-	private string? _plain;
 	private int? _hashCode;
 
 	/// <summary>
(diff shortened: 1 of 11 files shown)


cptjazz:
Turns out, my interpretation of "A memory block capable of holding at least minBufferSize elements of T." from the MemoryPool documentation is wrong.
But also, the logic before was working just by sheer luck, namely the default buffer size being sufficiently large.

I now rent 2x the segment length, based on the `Encode` method doubling the size in the worst case.

gregsdennis:
I'd like to see a test that fails to show that renting the segment size isn't enough, please.

cptjazz:
The first thing we can observe is, that the SpanExtension.Encode method can produce an output string that is worst case twice the size of the input string, namely when you pass something like "~~~" that turns into "~0~0~0".

So no matter where we get the buffer array from, we need to provide at least twice the segment length buffer to write the encoded version to.

The thin with MemoryPool.Rent() is, that when you pass an argument (like me at first passing in the segment length), you get a buffer that has _at least_ as many bytes available. But for small sizes like segment lengths of 10 you often get a buffer of 16 bytes back from the pool. If you happen to only have one or two encode-worthy characters in the segment you will not notice anything. So we must make sure to pass segment length times two as an argument here.

Your original version did not provide any arguments, leading the pool to return a buffer that is at least "default size" according to the docs -- what ever that means in practice. But given a sufficiently long segment with (or even without) encode-worthy characters, you will exceed the buffer.

It's hard to _reliably_ write a test that shows the error case, because it depends on what the pool gives you back as buffer; you could even get a 16k buffer when requesting only 3 bytes if the pool has no other free buffers to hand out.

Nevertheless I added tests for the positive case, which are always a good idea.

cptjazz:
Regarding the "... and why it only needs to happen here." part of your question: maybe there are more places that might need patching -- I will go and check.

cptjazz:
There are several cases of `MemoryPool`1.Shared.Rent()` where no minimum buffer size is set. Given sufficiently large inputs, all those buffers could be exceeded. 

I suggest we handle that in a new issue/PR.

gregsdennis:
> There are several cases of `MemoryPool<>.Shared.Rent()` where no minimum buffer size is set.

In my testing, it seemed that the minimum size was 512 items, whether I passed a minimum or not.  Because of this, I didn't submit a minimum everywhere because I couldn't fathom that someone would need more than 512 of something.  For example, I can't see someone needing more than 512 characters for a single pointer segment.

cptjazz:
If you pass an argument, say "8" it depends on what the pool has available. And that depends on the "history" of the pool; so the tests being run in a session and their order have an impact here. Remove the `* 2` from my change and run the tests I added a few times and they will fail half of the time.

Nevertheless, if you imply a size limit for the input, that should be communicated by a proper exception being thrown instead of a random IndexOutOfRangeException coming from the encoding logic.

gregsdennis:
Okay.  That makes sense.  .Net trying to be smart about things.

I kinda haphazardly stepped into this kind of memory management and tried to figure things out as I went along.  I'll graciously accept any improvements.

Regarding putting additional updates in other PRs, I prefer that `master` contains the latest released versions and not have any unreleased code.  But I also really only use mainline development here, which means that I'd want to update the version and release notes before merging these commits to `master`.

If you have other changes, I really don't mind having them in this PR, then merging everything at once.

cptjazz:
I will investigate, but it will take a few days ...

cptjazz:
BenchmarkDotNet v0.14.0, Windows 10 (10.0.19045.5247/22H2/2022Update)
Intel Core i7-6560U CPU 2.20GHz (Skylake), 1 CPU, 4 logical and 2 physical cores
.NET SDK 9.0.101
  [Host]   : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2
  .NET 8.0 : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2
  .NET 9.0 : .NET 9.0.0 (9.0.24.52809), X64 RyuJIT AVX2


| Method              | Job      | Runtime  | Count | Mean      | Error     | StdDev    | Median    | Ratio | RatioSD | Allocated | Alloc Ratio |
|-------------------- |--------- |--------- |------ |----------:|----------:|----------:|----------:|------:|--------:|----------:|------------:|
| PointerToString_New | .NET 8.0 | .NET 8.0 | 100   |  6.983 us | 0.1031 us | 0.1779 us |  6.953 us |  1.00 |    0.03 |         - |          NA |
| PointerToString_New | .NET 9.0 | .NET 9.0 | 100   |  8.147 us | 0.1599 us | 0.1496 us |  8.119 us |  1.17 |    0.04 |         - |          NA |
|                     |          |          |       |           |           |           |           |       |         |           |             |
| PointerToString_Old | .NET 8.0 | .NET 8.0 | 100   | 19.805 us | 0.3857 us | 0.4442 us | 19.789 us |  1.00 |    0.03 |         - |          NA |
| PointerToString_Old | .NET 9.0 | .NET 9.0 | 100   | 15.355 us | 0.2956 us | 0.4857 us | 15.128 us |  0.78 |    0.03 |         - |          NA |

I changed `ToString()` to require only one buffer, also I stackallocate the buffer if the pointer is small enough (< 1024 chars) which should account for all realistic scenarios.

I added buffer sizes for MemoryPool where easily feasible. I left a few open during parsing that would need some calculations. I want to improve parsing of pointers anyhow, but I will create another issue for that.

PTAL, @gregsdennis

cptjazz:
I added some optimizations for pointer parsing, as stated in #855.

BenchmarkDotNet v0.14.0, Windows 10 (10.0.19045.5247/22H2/2022Update)

11th Gen Intel Core i7-1185G7 3.00GHz, 1 CPU, 8 logical and 4 physical cores
.NET SDK 9.0.102
  [Host]   : .NET 8.0.12 (8.0.1224.60305), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  .NET 8.0 : .NET 8.0.12 (8.0.1224.60305), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  .NET 9.0 : .NET 9.0.1 (9.0.124.61010), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
 

| Method   | Job      | Runtime  | Count | Mean     | Error   | StdDev   | Median   | Gen0    | Allocated |
|--------- |--------- |--------- |------ |---------:|--------:|---------:|---------:|--------:|----------:|
| Parse    | .NET 8.0 | .NET 8.0 | 100   | 164.5 us | 2.91 us |  5.17 us | 162.7 us | 51.7578 | 317.97 KB |
| ParseOld | .NET 8.0 | .NET 8.0 | 100   | 312.4 us | 5.70 us |  9.98 us | 310.4 us | 99.1211 | 607.82 KB |
| Parse    | .NET 9.0 | .NET 9.0 | 100   | 153.3 us | 3.05 us |  8.46 us | 150.4 us | 48.0957 | 295.31 KB |
| ParseOld | .NET 9.0 | .NET 9.0 | 100   | 244.5 us | 4.88 us | 11.32 us | 244.5 us | 75.6836 | 464.85 KB |

cptjazz:
@gregsdennis, I'm done so far with the optimizations for JsonPointer.

src/JsonPointer/JsonPointer.cs
```diff
@@ -69,64 +67,183 @@ public class JsonPointer : IEquatable<JsonPointer>, IReadOnlyList<string>
...
+		if (!TryDecodeSegment(source, out segments[segmentIndex]))
+			return false;
+		
+		result = new JsonPointer(segments);
+		return true;
+	}
+
+#endif
+
+	private static bool TryDecodeSegment(ReadOnlySpan<char> encoded, out string? result)
```
gregsdennis, line 187: Please add nullability annotation, and please put back into the `SpanExtensions` class.  I think it reads better that way.

src/JsonPointer/JsonPointer.cs
```diff
@@ -28,7 +27,6 @@ public class JsonPointer : IEquatable<JsonPointer>, IReadOnlyList<string>
 	public static readonly JsonPointer Empty = new();
 
 	private readonly string[] _decodedSegments;
-	private string? _plain;
```
gregsdennis, line 31: Surely it would help to cache the string rather than rebuilding it each time.  This would be populated either on `.Parse()` (the original string input) or lazily on `.ToString()` (if the pointer was created with `.Create()`).

gregsdennis (changes requested):
Looks really good.  Just a few things.

gregsdennis:
I've separately merged a better test reporter (in addition to the summary that is posted to the PR).  This should help find failures.

src/JsonPointer/JsonPointer.cs
cptjazz: I brought back the caching and populate it in situations where this is possible without allocations. In all other cases, the cost is moved to `ToString()`

cptjazz:
It's hard for me to run the _whole_ test suite locally, because several tests always fail(ed).
I fixed the culture for some of them now (last commit), because they assumed English output which is not true on my German OS.

gregsdennis:
Ah, yeah, those tests probably weren't an issue in Germany before the recent German error translations were added.

gregsdennis:
Everything passes for me locally.  Might be one of the flaky datagen tests.

gregsdennis (approved):

gregsdennis:
Thanks @cptjazz!

PR #857, opened 2025-01-15
fix #856; allow enum serialization with duplicated member values

### LINKED ISSUE #856: `EnumStringConverter` fails on Enum serialization when multiple members have same value
### Nuget Package

Json.More.Net

### Package Version

2.1.0

### Operating System

Windows

.Net (5 or after)

`EnumStringConverter` fails on Enum serialization when multiple members have same value.

This could be reproduced by Serializing `System.Net.HttpStatusCode` which has both members (Ambiguous, MultipleChoices) assigned to same value (300).

### What did you expect?

While this is really edge case scenario, I would say that since C# supports such enum declaration, `EnumStringConverter` should conform to it.

When serializing, I would expect that it should rely on member name or member `DescriptionAttribute` without value.
When deserializing, I assume library doesnt support parsing number values, therefore no changes required.

`JsonSerializer.Serialize(HttpStatusCode.BadRequest, new JsonSerializerOptions { Converters = { new EnumStringConverter<HttpStatusCode>() } });`

Converter fails in `EnsureMap()` method since it tries to create `write cache` dictionary based on enum values, which does not support duplicates. 

nazar-kuzo:
### Description

Allow enum serialization with duplicated member values.

This does not affect the public API surface;
The changes does not cause breaks in either developer experience or behavior;

### Links

Resolves  #856

### Checks

### DIFF
--- a/src/Json.More/EnumStringConverter.cs
+++ b/src/Json.More/EnumStringConverter.cs
@@ -177,16 +177,29 @@ public class EnumStringConverter<[DynamicallyAccessedMembers(DynamicallyAccessed
 		{
 			if (_readValues != null && _writeValues != null) return;
 
-			var map = typeof(T).GetFields()
-				.Where(f => !f.IsSpecialName)
-				.Select(f => new
-				{
-					Value = (T)Enum.Parse(typeof(T), f.Name),
-					Description = f.GetCustomAttribute<DescriptionAttribute>()?.Description ?? f.Name
-				})
-				.ToList();
-			_readValues = map.ToDictionary(v => v.Description, v => v.Value);
-			_writeValues = map.ToDictionary(v => v.Value, v => v.Description);
+			var fields = typeof(T).GetFields();
+
+			var readValues = new Dictionary<string, T>(capacity: fields.Length);
+			var writeValues = new Dictionary<T, string>(capacity: fields.Length);
+
+			foreach (var field in fields.Where(f => !f.IsSpecialName))
+			{
+				var value = (T) Enum.Parse(typeof(T), field.Name);
+				var description = field.GetCustomAttribute<DescriptionAttribute>()?.Description ?? field.Name;
+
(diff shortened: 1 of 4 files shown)


nazar-kuzo:
@dotnet-policy-service agree

src/Json.More/EnumStringConverter.cs
```diff
@@ -177,16 +177,29 @@ private static void EnsureMap()
...
+
+#if NET8_0_OR_GREATER
+				readValues.TryAdd(description, value);
+				writeValues.TryAdd(value, description);
+#else
+				if (!readValues.ContainsKey(description))
+					readValues.Add(description, value);
+				if (!writeValues.ContainsKey(value))
+					writeValues.Add(value, description);
+#endif
```
gregsdennis, line 198: Instead of having the compiler directives here, I'd prefer having a `NetCompatibilityExtensions` that defines a `.TryAdd()` method for the .Net versions that are missing it.  I think it makes the usage site cleaner and easier to read.

nazar-kuzo: Thanks for the idea. In that regard I decided to opt out of using `TryAdd` since under the hood it is simple `ContainsKey/Add` method calls, no optimizations, so vanilla approach fits much better

gregsdennis (approved):
Thanks for this.

Please bump the patch version in the project file and add a release note.  For the release note, follow the pattern in other notes.  This PR is linked in the header, and the issue is linked in the note text.  Feel free to add a thanks to yourself. 😄

gregsdennis:
@nazar-kuzo just a couple housekeeping items here, then it can be merged.

nazar-kuzo:
@gregsdennis sorry for the delay. Updated release notes and bumped version

gregsdennis:
The failure seems to be a permissions issue.  The changes look good.  Thanks!

PR #864, opened 2025-01-25
RelativeJsonPointer: use span APIs when on .NET8+

cptjazz:
### Description

Use modern Span-based APIs to avoid allocations when parsing RelativeJsonPointer

### Links

Resolves #863 
Depends on #868 

### Checks

- [x] I have read and understand the [Code of Conduct](https://github.com/json-everything/json-everything/blob/master/CODE_OF_CONDUCT.md) and [Contribution Guidelines](https://github.com/json-everything/json-everything/blob/master/CONTRIBUTING.md).

PR #870, opened 2025-02-03
Multi release

### LINKED ISSUE #852: Format 'hostname' does not accept hostnames longer than 24 characters when using RequireFormatValidation
### Nuget Package

JsonSchema.Net

### Package Version

7.3.1

### Operating System

Windows

None

Validating hostnames against a schema that specifies `"format": "hostname"` on a `"type": "string"` does not accept hostnames longer than 24 characters.

### What did you expect?

I expected to be able to validate hostnames up to 255 characters as per RFC-1123. It seems this package is using the (old) RFC-952 spec.

RFC-1123:
> Host software MUST handle host names of up to 63 characters and
      SHOULD handle host names of up to 255 characters.

```
    [Fact]
    public void HostNames_Exceeding24Characters_AreValid()
    {
        var schema = "{\"type\": \"string\", \"format\": \"hostname\"}";
        var jsonSchema = JsonSerializer.Deserialize<JsonSchema>(schema);
        Assert.NotNull(jsonSchema);

        var json = JsonSerializer.SerializeToNode("hostname.exceeding24characte.rs");
        var results = jsonSchema.Evaluate(json, new EvaluationOptions { RequireFormatValidation = true });

        Assert.True(results.IsValid);
    }
```

_No response_

### LINKED ISSUE #863: Avoid allocations in RelativeJsonPointer
### Nuget Package

JsonPointer.Net

### Package Version

5.3.0

When parsing RelativeJsonPointer, we can use Span-based APIs of NET8/9 to avoid several `ToString()` calls, like that:

```
#if NET8_0_OR_GREATER
		var parentSteps = uint.Parse(span[..i]);
#else
		var parentSteps = uint.Parse(span[..i].ToString());
#endif
```

_No response_

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### LINKED ISSUE #867: .NET9 assemblies missing in NuGet package
### Nuget Package

JsonSchema.Net

### Package Version

7.3.1

### Operating System

None

None

The project targets .NET9, but the actual assemblies are missing from the NuGet package. 

### What did you expect?

.NET9 assemblies being in the package.

Can be checked by downloading from https://www.nuget.org/packages/JsonSchema.Net and unzipping the package.

_No response_

gregsdennis:
### Description

- Updates `hostname` and `idn-hostname` formats for JsonSchema.Net
- Updates JsonSchema.Net nuget package to include libs for .Net 9
- Further reduces allocations in JsonPointer.Net

### Links

Resolves #852
Resolves #863
Resolves #867 

Depends on #864

### Checks

### DIFF
--- a/.github/workflows/publish-schema-sign.yml
+++ b/.github/workflows/publish-schema-sign.yml
@@ -129,7 +129,7 @@ jobs:
 
     # Install the code signing tool    
     - name: Install Sign CLI tool
-      run: dotnet tool install --tool-path . sign --version 0.9.1-beta.24529.1
+      run: dotnet tool install --tool-path . --prerelease sign
     
     # Login to Azure using a ServicePrincipal configured to authenticate against a GitHub Action
     - name: 'Az CLI login'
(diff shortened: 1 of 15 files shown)


gregsdennis:
So it looks like _nuget.exe_ isn't supported on Ubuntu anymore since it requires mono.  My schema lib and its language packs aren't supported by `dotnet pack`, and so now I have to figure out how to do this differently. 

I'd prefer not to run the build on windows, but it might be the easiest stop-gap.

glennawatson:
If you're doing Linux building look into dotnet pack command. It works over Multiplatform

gregsdennis:
It does work multi-platform, but it doesn't do what I need it to do. Specifically, it doesn't support nuspec files.

glennawatson:
> @glennawatson Please see my question on SO: https://stackoverflow.com/q/79410903/878701

See  
'''
    <IncludeBuildOutput>false</IncludeBuildOutput>
'''

glennawatson:
Then you can include your DLL directly

https://github.com/reactiveui/ReactiveUI.SourceGenerators/blob/main/src/ReactiveUI.SourceGenerators/ReactiveUI.SourceGenerators.csproj

gregsdennis:
I've been able to get that working except for one thing: I can't specify dependencies for the language packs.  It always takes the dependencies for the main lib.

I still need to update the build script itself.

cptjazz:
Maybe the easiest solution is to create a csproj per-language, set a project dependency to JsonSchema itself and let plain dotnet pack do the rest.

I know, this solution has its own drawbacks with 10+ new projects in the solution, but otherwise you will always work against the system to achieve the packaging layout you want.

gregsdennis:
I tried that. The problem is that it builds the language assembly with a separate internal assembly ID and the main library can load its resources as its own. 

What I have works, and I'm satisfied with it. I still need to go through and update the other builds, but I'll do that after the package signing work is done.

Issue #558, opened 2023-11-08
 Validating against a schema with an oneOf condition using the date and date-time format results in a invalid validation


FelixJungensbluth:
### Nuget Package

JsonSchema.Net

### Package Version

5.3.1

### Operating System

Windows

### .Net Target (if relevant, please specify the version in the description)

.Net (5 or after)

### A clear and concise description of the bug.

Validating against a schema with an oneOf condition using the date and date-time format with the value like 2019-07-03T13:13:13 results in a invalid validation with the error: 
```
"errors": {
        "oneOf": "Expected 1 matching subschema but found 2"
      },
```

Schema:
```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "description": "A schema to validate fields that can either be a date or a date-time",
  "properties": {
    "dateCreated": {
      "oneOf": [
        {
          "type": "string",
          "description": "A date string",
          "format": "date"
        },
        {
          "type": "string",
          "description": "A date-time string",
          "format": "date-time"
        }
      ]
    }
  },
  "required": ["dateCreated"]
}

```

### What did you expect?

valid = true

### Please add test code or steps to reproduce the behavior.

![oneOf-validation-error](https://github.com/gregsdennis/json-everything/assets/49518120/d80d79d0-9e34-4c4a-b3c4-6a89fb48b2de)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gregsdennis:
As I mention in my [docs](https://docs.json-everything.net/schema/basics/#schema-format), I make an effort to validate `format`s closely, but I make no guarantee of them working 100%.

I recommend that you create and register your own implementation of these formats to suit your needs.  I'll have a look at improving them in the future.  Also a PR would be welcome.

FelixJungensbluth:
Thanks for your quick response. I'll create my own implementations for these formats, as you recommended, and create a PR afterwards.

PR #882, opened 2025-03-17
Changed Date and Time validation to not validate against the DateTime…

### LINKED ISSUE #881: Validating DateTime value against oneOf containing both "date" and "date-time" leads to subschema missmatch
### Nuget Package

JsonSchema.Net

### Package Version

7.3.3

### Operating System

Windows

None

We are having the exact same issue as #558.
I decided to take a closer look as to why the error comes up and generate a PR if I found something.

I found that the [`CheckDate()`](https://github.com/json-everything/json-everything/blob/2ff12adc80baae018147f453a016358182ab1296/src/JsonSchema/Formats.cs#L300) method calls the [`CheckDateFormat()`](https://github.com/json-everything/json-everything/blob/2ff12adc80baae018147f453a016358182ab1296/src/JsonSchema/Formats.cs#L315C22-L315C37) with a fixed format of "yyyy-MM-dd".
The `CheckDateFormat()` method does the `DateTime.TryParseExact()` call when a format is provided.
However, it always does a RegEx check against the DateTime-Regex if the previous is either skipped or returned as false.

This means that both `CheckDate` and `CheckTime` will return `true` if the provided input is in a full DateTime format, which does not seem right.

I will create a PR with a correction suggestion, however this might be a breaking change since old validations might fail now.

### What did you expect?

A valid validation result

Schema used for validation:
```
{
    "$schema": "https://json-schema.org/draft/2019-09/schema",
    "description": "A schema to validate fields that can either be a date or a date-time",
    "properties": {
        "dateCreated": {
            "title": "Date when it was created",
            "type": "string",
            "oneOf": [
                {
                    "type": "string",
                    "format": "date"
                },
                {
                    "type": "string",
                    "format": "date-time"
                }
            ]
        }
    },
    "required": [
        "dateCreated"
    ]
}
```

Json used for testing:
```
{
  "dateCreated": "2025-01-02T23:03:22.222Z"
}
```

I would also suggest that the JSON-Schema-Test-Suite can be extended to include cases where the date format is given a date time value, etc. to cover these cases and ensure that all libraries have the correct validation.

Dragonsangel:
### Description
Changed Date and Time validation to not validate against the DateTime RegEx format.

The added unit tests could be removed if the cases are added to the JSON-Schema-Test-Suite.
Let me know if I should add them there and remove them from this PR.

These changes _could_ break existing implementations due to the fact that `date` and `time` formats would historically have a valid validation, which with these changes will change to an invalid validation.

### Links

This should resolve #881 and is related to #558

### Checks

- [ x ] I have read and understand the [Code of Conduct](https://github.com/json-everything/json-everything/blob/master/CODE_OF_CONDUCT.md) and [Contribution Guidelines](https://github.com/json-everything/json-everything/blob/master/CONTRIBUTING.md).

### DIFF
--- a/src/JsonSchema.Tests/GithubTests.cs
+++ b/src/JsonSchema.Tests/GithubTests.cs
@@ -961,6 +961,108 @@ public class GithubTests
 	}
 
 	[Test]
+	public void Issue881_DateTimeValidationPassesWithOneOfSchema()
+	{
+		var file = GetFile(881, "schema");
+
+		var schema = JsonSchema.FromFile(file);
+
+		var instance = new JsonObject()
+		{
+				new KeyValuePair<string, JsonNode?>("oneOfDates", "2025-01-02T23:03:22.222Z")
+		};
+
+		var result = schema.Evaluate(instance, new EvaluationOptions { OutputFormat = OutputFormat.List });
+
+		result.AssertValid();
+	}
+
+	[Test]
+	public void Issue881_DateValidationPassesWithOneOfSchema()
+	{
+		var file = GetFile(881, "schema");
+
+		var schema = JsonSchema.FromFile(file);
+
+		var instance = new JsonObject()
+		{
+				new KeyValuePair<string, JsonNode?>("oneOfDates", "2025-01-02")
+		};
+
+		var result = schema.Evaluate(instance, new EvaluationOptions { OutputFormat = OutputFormat.List });
+
+		result.AssertValid();
+	}
+
+	[Test]
+	public void Issue881_DateValidationFailsAgainstDateTime()
+	{
+		var file = GetFile(881, "schema");
+
+		var schema = JsonSchema.FromFile(file);
+
+		var instance = new JsonObject()
+		{
+				new KeyValuePair<string, JsonNode?>("dateWithDateFormat", "2025-01-02T23:03:22.222Z")
+		};
+
+		var result = schema.Evaluate(instance, new EvaluationOptions { OutputFormat = OutputFormat.List });
+
+		result.AssertInvalid();
+	}
+
+	[Test]
+	public void Issue881_DateValidationPassesAgainstDate()
+	{
+		var file = GetFile(881, "schema");
+
+		var schema = JsonSchema.FromFile(file);
+
+		var instance = new JsonObject()
+		{
+				new KeyValuePair<string, JsonNode?>("dateWithDateFormat", "2025-01-02")
+		};
+
+		var result = schema.Evaluate(instance, new EvaluationOptions { OutputFormat = OutputFormat.List });
+
+		result.AssertValid();
+	}
+
+	[Test]
+	public void Issue881_TimeValidationFailsAgainstDateTime()
+	{
+		var file = GetFile(881, "schema");
+
+		var schema = JsonSchema.FromFile(file);
+
+		var instance = new JsonObject()
+		{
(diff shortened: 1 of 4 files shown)


gregsdennis:
This looks really good.  Thanks for the help.

Please update the [patch version](https://github.com/json-everything/json-everything/blob/2ff12adc80baae018147f453a016358182ab1296/src/JsonSchema/JsonSchema.csproj#L20) so a new package will publish when merged.

> Let me know if I should add them there and remove them from this PR.

There's no harm in having them in both places.

> These changes could break existing implementations due to the fact that date and time formats would historically have a valid validation.

This is a fix correcting behavior, so I wouldn't consider it breaking.  We shouldn't rely on incorrect behavior.

src/JsonSchema.Tests/GithubTests.cs
```diff
@@ -960,6 +960,108 @@ public void Issue600_BaseDocumentOutputSchemaLocation()
...
+		var schema = JsonSchema.FromFile(file);
+
+		var instance = new JsonObject()
+		{
+				new KeyValuePair<string, JsonNode?>("dateWithDateFormat", "2025-01-02T23:03:22.222Z")
+		};
+
+		var result = schema.Evaluate(instance, new EvaluationOptions { OutputFormat = OutputFormat.List });
+
+		result.AssertInvalid();
```
gregsdennis, line 1011: Something isn't right here. The evaluation options need to have `RequireFormatValidation` set in order to process `format`.  I'd expect this test to fail.

https://docs.json-everything.net/schema/basics/#schema-format

https://docs.json-everything.net/schema/basics/#schema-options

This option needs to be set on all of these tests.

src/JsonSchema.Tests/GithubTests.cs
```diff
@@ -960,6 +960,108 @@ public void Issue600_BaseDocumentOutputSchemaLocation()
...
+	[Test]
+	public void Issue881_DateValidationFailsAgainstDateTime()
+	{
+		var file = GetFile(881, "schema");
+
+		var schema = JsonSchema.FromFile(file);
+
+		var instance = new JsonObject()
+		{
+				new KeyValuePair<string, JsonNode?>("dateWithDateFormat", "2025-01-02T23:03:22.222Z")
```
gregsdennis, line 1006:
```suggestion
				["dateWithDateFormat"] = "2025-01-02T23:03:22.222Z"
```

Please update all of these.

gregsdennis (changes requested):

src/JsonSchema.Tests/GithubTests.cs
Dragonsangel: Interestingly that the tests don't fail. I stepped through the tests and all of them execute the relevant CheckDate/CheckTime/CheckDateTime methods.
I'll change it as requested, might however need to check why it behaves differently than expected.

gregsdennis: I'd debug it myself, but unfortunately my computer is in the shop right now.  Some kind of hardware problem.

Dragonsangel: I debugged this and from what I understand is that the validation of Schema known Format Keywords is enforced, regardless of the `RequireFormatValidation` setting since the validator understands and knows the format.

The logic for this is within the [FormatKeyword.cs](https://github.com/json-everything/json-everything/blob/2ff12adc80baae018147f453a016358182ab1296/src/JsonSchema/FormatKeyword.cs#L74) class

gregsdennis:
Thanks for this.  Once the build passes, I'll bring it in, and the build should push an update automatically.

PR #887, opened 2025-03-25
Fix producing JSON schema with commented enums

### LINKED ISSUE #886: JsonSchemaBuilder might build incomplete JSON schemas (missing $defs)
### Nuget Package

JsonSchema.Net.Generation

### Package Version

5.0.0

### Operating System

None

None

When generating a JSON schema with `JsonSchemaBuilder` and reading comments from an xmldoc file, the generated JSON schema might be incomplete, i.e. missing the `$defs` part.

### What did you expect?

Running the sample code (from next paragraph) should produce the following JSON schema.

```json
{
  "type": "object",
  "properties": {
    "Size": {
      "$ref": "#/$defs/size",
      "description": "The size of the pizza."
    }
  },
  "$defs": {
    "size": {
      "description": "All possible sizes for a pizza.",
      "enum": [
        "Small",
        "Medium",
        "Large"
      ]
    }
  }
}
```

But using `JsonSchema.Net.Generation` 5.0.0 it produces this incomplete JSON schema. Note how the `$defs` is completely missing.

```json
{
  "type": "object",
  "properties": {
    "Size": {
      "$ref": "#/$defs/size",
      "description": "The size of the pizza."
    }
  }
}
```

**JsonSchemaEnumBug.csproj**

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net9.0</TargetFramework>
    <Nullable>enable</Nullable>
    <GenerateDocumentationFile>true</GenerateDocumentationFile>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="JsonSchema.Net.Generation" Version="5.0.0" />
  </ItemGroup>

</Project>
```

**Program.cs**

```csharp
using System;
using System.Text.Json;
using Json.Schema;
using Json.Schema.Generation;
using Models;

var config = new SchemaGeneratorConfiguration();
config.RegisterXmlCommentFile<Pizza>("JsonSchemaEnumBug.xml");
var schema = new JsonSchemaBuilder().FromType<Pizza>(config).Build();
var serializerOptions = new JsonSerializerOptions { WriteIndented = true };
Console.WriteLine(JsonSerializer.Serialize(schema, serializerOptions));

namespace Models
{
    public class Pizza
    {
        /// <summary>
        /// The size of the pizza.
        /// </summary>
        public Size Size { get; init; }
    }

    /// <summary>
    /// All possible sizes for a pizza.
    /// </summary>
    public enum Size
    {
        Small,
        Medium,
        Large,
    }
}
```

Save those two files in the same directory then type `dotnet run` to see the result.

I think I have identified the root cause and will submit a pull request to address this issue soon.

0xced:
### Description

This pull request fixes producing JSON schema with commented enums as described in #886.

A new test `Issue886_CommentedEnum` was added to ensure that the JSON schema is produced as expected.

Note that we can do `schema.References[0]` because the `toReintegrate` schema is ensured to have a single reference just above: `var toReintegrate = Cache.Where(x => (x.Value.References.Count == 1 && ...`

This pull request does **not** affect the public API surface.
This pull request does **not** break either developer experience or behavior.

### Links

Resolves #886

### Checks


src/JsonSchema.Generation/SchemaGenerationContextCache.cs
```diff
@@ -46,16 +46,15 @@ internal static TypeGenerationContext GetRoot(Type type)
 
 		foreach (var schema in toReintegrate)
 		{
-			foreach (var context in schema.References)
-			{
-				var contextKeywords = context.Intents.Select(x => x.GetType());
-				var schemaKeywords = schema.Intents.Select(x => x.GetType());
-				if (contextKeywords.Intersect(schemaKeywords).Except([typeof(RefIntent)]).Any()) continue;
+			var context = schema.References[0];
```
gregsdennis, line 49: This should fail in cases where there are more than one reference.  I might have a test for that, but in case not, can you add a test where there is more than one reference to type?

I also fail to see how this fixes the problem.

gregsdennis:
I see that you added the test first.  That's great, however the PR wasn't created with just the failing test, so the actions never ran against that commit.  The reason I ask to put the failing tests in an initial commit is so that I can see on the PR that they're failing.

src/JsonSchema.Generation/SchemaGenerationContextCache.cs
0xced: I edited my initial post with this comment:

> Note that we can do schema.References[0] because the toReintegrate schema is ensured to have a single reference just above: var toReintegrate = Cache.Where(x => (**x.Value.References.Count == 1** && ...

I hesitated to edit the post or add a new comment in order not to spam you with an e-mail notification. 😄

> I also fail to see how this fixes the problem.

There were two `foreach` loops with an `if (...) continue;` skip condition. Then between the two `foreach` loops there's `schema.References.Clear();` but this was **not** meant to be called when the schema is not reintegrated (i.e. skipped by the `continue`). By removing the inner `foreach` loop which is not an actual loop since we've ensured that the schema has a single reference, `schema.References.Clear();` is only called when the schema has actually been reintegrated and not when it was skipped.

0xced:
I just force-pushed the first commit (e57b98051d8c592a47015df8b1f1877600dd5b69) in order to trigger a [run](https://github.com/json-everything/json-everything/actions/runs/14101180186) but then this happened: https://github.com/json-everything/json-everything/runs/39497668044#r8s7

Simply adding `Issue886_CommentedEnum` without touching any code in the library made another test fail!
> ❌ Issue646_PropertyWithFalse
> 	  Assert.That(results.IsValid, Is.True)
> 	  Expected: True
> 	  But was:  False

I don't understand how this is possible. 🤯

Anyway, here's how the test is supposed to fail before the fix in the second commit:
```
Expected: {"type":"object","properties":{"Size":{"$ref":"#/$defs/pizzaSizeInClientTests","description":"The size of the pizza."}},"$defs":{"pizzaSizeInClientTests":{"description":"All possible sizes for a pizza.","enum":["Small","Medium","Large"]}}}
Actual:   {"type":"object","properties":{"Size":{"$ref":"#/$defs/pizzaSizeInClientTests","description":"The size of the pizza."}}}
   at TestHelpers.JsonAssert.AreEquivalent(JsonNode expected, JsonNode actual) in ~/json-everything/src/TestHelpers/JsonAssert.cs:line 19
   at Json.Schema.Generation.Tests.AssertionExtensions.AssertEqual(JsonSchema expected, JsonSchema actual) in ~/json-everything/src/JsonSchema.Generation.Tests/AssertionExtensions.cs:line 19
   at Json.Schema.Generation.Tests.ClientTests.Issue886_CommentedEnum() in ~/json-everything/src/JsonSchema.Generation.Tests/ClientTests.cs:line 545
```

src/JsonSchema.Generation/SchemaGenerationContextCache.cs
gregsdennis: Okay. I'll need to look more wholesale when I get my computer back from the shop.  I'm down to my work computer and I don't do my OSS projects on that.

0xced:
I edited the pull request title, content and commits (force pushed) to mention "commented types" instead of "commented enums" to be clear that this issue is not about enums but about comments.

0xced:
So I just force-pushed commit a088762950f965d7c4ffd76b95fc1aa1c62193f2 (Add new test that covers commented types) and this [run](https://github.com/json-everything/json-everything/actions/runs/14101620633/job/39498865672?pr=887) started. This is supposed to fail but all tests are ✅. Looking at the **Run actions/checkout@v4** step, it checked out commit be57a968e9a136002606f09f16c53e10ef33f27d which is the current `master`!

It's looks like there's something fishy with the GitHub actions. 🤔

0xced:
This @github-actions comment is also weird:
> This pull request **removes** 25 and **adds** 6 tests. _Note that renamed tests count towards both._

This pull request actually only adds a single test and removes and renames none!

gregsdennis:
Yeah, that's broken. I don't know what's going on and haven't looked into it. It's someone else's action.

src/JsonSchema.Generation/SchemaGenerationContextCache.cs
0xced: Did you have a chance to take a look at it on your personal computer?

0xced:
Hey Greg,
Just a quick reminder to take a look at that pull request when you get a chance. No rush, I just wanted to make sure it didn’t slip through the cracks. Let me know if you have any questions or need anything from me.

Thanks!

gregsdennis:
Okay.  The change looks good.  Please make the following additional edits:

1. Bump the patch version of the library (in the .csproj).
2. Add an entry in the [release notes](https://github.com/json-everything/json-everything/blob/master/tools/ApiDocsGenerator/release-notes/rn-json-schema-generation.md) for the fix.  Feel free to give yourself credit.

Then I'll merge this in and a new package will publish.

0xced:
I just rebased the branch onto `master`, bumped the version to 5.0.1 and added release notes. Hope it's in good shape for a new release. 🚀

gregsdennis:
Thanks. It'll publish soon.

PR #901, opened 2025-05-24
JsonPath and JsonE equivalency to JsonNode when dealing with CLR types that JsonNode equates to "String kind"

### LINKED ISSUE #900: JsonPath incorrectly evaluates against CLR types that JsonNode deems to be of "String kind"
### Nuget Package

JsonPath.Net

### Package Version

2.1.1

### Operating System

None

.Net (5 or after)

JsonNode treats several CLR types as "String Kind".  This includes (but isn't limited to) DateTime, DateTimeOffset, Timespan, DateOnly and TimeOnly.

The JsonPath implementation is unaware of these CLR "String kind" equivalents and therefore does not return the expected results.

### What did you expect?

JsonPath has the same notion of "String kind" as JsonNode and therefore `JsonPath.Evaluate` returns nodes that meet the criteria using the string representation of these CLR types. 

Given:
`
			var root = new JsonObject
			{
				["foo"] = new JsonObject
				{
					["bar"] = (JsonValue)new DateTime(2025, 1, 2, 3,4,5,6,7)
				}
			};
`
Expect:
`
Assert.That(JsonPath.Parse("$[?@.bar == \"2025-01-02T03:04:05.006007\"]").Evaluate(root).Matches.Count, Is.EqualTo(1));
`

Given:
`
var root = new JsonObject { ["foo"] = new JsonObject { ["bar"] = JsonValue.Create(new DateOnly(2025, 1, 1)) } };
`
Expect:
`
Assert.That(JsonPath.Parse("$[?@.bar == \"2025-01-01\"]").Evaluate(root).Matches.Count, Is.EqualTo(1));
`

Looking at the code I can see that `EqualToOperator` and `NotEqualToOperator` uses `Json.More.JsonNodeExtensions.IsEquivalentTo` which in turn uses `Json.More.JsonNodeExtensions.GetString` but the `LessThanOperator` (et al) try for `string` directly.

I propose that `Json.More.JsonNodeExtensions.GetString` is enhanced to be aware of other "String Kind" equivalent types and that `LessThanOperator` (et al) switch to using  `Json.More.JsonNodeExtensions.GetString` 

I also see that JsonE also has operators that use the same pattern as JsonPath and the same modification should likely be applied to JsonE as well.

BryantL:
### Description

Implements a minimal change to JsonPath and JsonE to allow equivalency in "String kind" handling to that implemented in JsonNode.
Public interface is not modified however the behaviours of JsonPath and JsonE is changed when evaluating a JsonNode DOM that contains "String kind" equivalent CLR types.

### Links

Resolves #900     

### Checks

- [X] I have read and understand the [Code of Conduct](https://github.com/json-everything/json-everything/blob/master/CODE_OF_CONDUCT.md) and [Contribution Guidelines](https://github.com/json-everything/json-everything/blob/master/CONTRIBUTING.md).

### DIFF
--- a/src/Json.More/JsonNodeExtensions.cs
+++ b/src/Json.More/JsonNodeExtensions.cs
@@ -223,7 +223,9 @@ public static class JsonNodeExtensions
 		if (value.TryGetValue(out string? s)) return s;
 		if (value.TryGetValue(out char c)) return c.ToString();
 
-		return null;
+		return value.GetValueKind() == JsonValueKind.String
+			? value.ToJsonString()[1..^1] //strip JSON literal double quotes
+			: null;
 	}
 
 	/// <summary>
--- a/src/JsonE/Expressions/Operators/GreaterThanOperator.cs
+++ b/src/JsonE/Expressions/Operators/GreaterThanOperator.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Text.Json;
 using System.Text.Json.Nodes;
 using Json.More;
 
@@ -14,9 +15,9 @@ internal class GreaterThanOperator : IBinaryOperator
 			right is not JsonValue rValue)
 			throw new InterpreterException("infix: > expects numbers/strings > numbers/strings");
 
-		if (lValue.TryGetValue(out string? leftString) &&
-			rValue.TryGetValue(out string? rightString))
-			return string.Compare(leftString, rightString, StringComparison.Ordinal) > 0;
+		if (lValue.GetValueKind() == JsonValueKind.String &&
+		    rValue.GetValueKind() == JsonValueKind.String)
+			return string.Compare(lValue.GetString(), rValue.GetString(), StringComparison.Ordinal) > 0;
 
 		var lNumber = lValue.GetNumber() ?? throw new InterpreterException("infix: > expects numbers/strings > numbers/strings");
 		var rNumber = rValue.GetNumber() ?? throw new InterpreterException("infix: > expects numbers/strings > numbers/strings");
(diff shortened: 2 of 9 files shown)


BryantL:
@dotnet-policy-service agree

BryantL:
In `JsonNodeExtensions.GetString` the minimal change implementation means that we call `ToJsonString` and then strip the double quotes (transforming from string literal back to string).
A potentially more performant way to do this would be to use Utf8Writer such that the double quotes could be stripped without incurring the double string create.  However using Utf8Writer in the most performant way likely means pooling of Utf8Writers which adds complexity.

Happy to look at this as an option if its deemed worth it.

src/Json.More/JsonNodeExtensions.cs
```diff
@@ -223,7 +223,9 @@ public static string AsJsonString(this JsonNode? node, JsonSerializerOptions? op
 		if (value.TryGetValue(out string? s)) return s;
 		if (value.TryGetValue(out char c)) return c.ToString();
 
-		return null;
+		return value.GetValueKind() == JsonValueKind.String
+			? value.ToJsonString()[1..^1] //strip JSON literal double quotes
```
gregsdennis, line 227: This is going to break for strings that contain characters that are JSON-encoded.

BryantL: Given that stored strings are dealt with on line 223, this is only going to deal with clr types that JsonNode considers to be strings and need converting. 
None of these do any encoding.
I guess in future it is possible that JsonNode could consider a clr type to be a string that does need encoding but I doubt it since it would be hard to justify a string equivalent type that needs an encoding

PR #910, opened 2025-06-28
add test to replicate behavior

### LINKED ISSUE #906: SchemaGenerationContextCache to dictionary issue
### Nuget Package

json-everything.net (playground)

### Package Version

7.3.4

### Operating System

None

None

I found an issue when creating the schema for a type that contains both a property of type `T[]` and some other property of type `IEnumerable<T>`. There is a ToDictionary() call here https://github.com/json-everything/json-everything/blob/master/src/JsonSchema.Generation/SchemaGenerationContextCache.cs#L65, and for the case I mentioned both DefinitionNames have the same value arrayOfT and ToDictionary() throws.

### What did you expect?

For starters, not to crash 😄 . I'm not sure about aliasing `IEnumerable<T>` and T[] though in the output... I suggest keeping them separate.

I encountered the issue when attempting to create schema for the .net DataTable type, but it can be simplified to the case I described.

_No response_

gregsdennis:
### Description

Adds a qualifier to ensure that multiple types that produce the same schema (and schema definition name) do not cause a collision when producing the definitions list.

### Links

Resolves #906     

### Checks


gregsdennis:
... what?  That should have failed.  It also doesn't say that a test was added.

PR #963, opened 2025-12-13
`PatchExtensions` refactor for efficiency and style

stephenmk:
### Description

These changes remove the `System.Linq` dependency in the class `PatchExtensions`. These changes are purely for style and efficiency; there are no API changes or behavioral changes.

More details are included in the expanded commit messages.

### Links

None

### Checks

- [x] I have read and understand the [Code of Conduct](https://github.com/json-everything/json-everything/blob/master/CODE_OF_CONDUCT.md) and [Contribution Guidelines](https://github.com/json-everything/json-everything/blob/master/CONTRIBUTING.md).

PR #970, opened 2025-12-19
Optimize array patches when original array length is zero

### LINKED ISSUE #964: `PatchExtensions.PatchForArray` can produce shorter patches when original array is empty
### Nuget Package

JsonPatch.Net

### Package Version

4.0.0

I did not include this in #963 because this is a behavioral change rather than a simple runtime performance optimization.

If we want to create a patch by comparing two JSON arrays, `A = []` and `B = [0, 1, 2, ..., 99]`, the current `CreatePatch` implementation in package 4.0.0 produces a patch with 100 operations. Since `A` is empty, we could optimize this by using a single `Replace` operation instead.

```csharp
	/// This test will pass in version 4.0.0
	[Test]
	public void CreatePatch_AddWholeArray()
	{
		var largeNumberArray = Enumerable.Range(start: 0, count: 100).ToArray();
		var model = new TestModel
		{
			Numbers = []
		};
		var model2 = new TestModel
		{
			Numbers = largeNumberArray
		};
		var patch = model.CreatePatch(model2, TestEnvironment.SerializerOptions);
		var final = patch.Apply(model, TestEnvironment.SerializerOptions);
		Assert.That(final!.Numbers, Is.EqualTo(largeNumberArray));
	}

	/// This test does not pass in version 4.0.0; the patch will have 100 operations.
	[Test]
	public void CreatePatch_OptimalAddWholeArrayPatchSize()
	{
		var largeNumberArray = Enumerable.Range(start: 0, count: 100).ToArray();
		var original = new TestModel
		{
			Numbers = []
		};
		var target = new TestModel
		{
			Numbers = largeNumberArray
		};
		var patch = original.CreatePatch(target, TestEnvironment.SerializerOptions);
		Assert.That(patch.Operations, Has.Count.EqualTo(1));
	}
```

We can get a shorter patch by updating the `PatchForArray` method to check if either the original array is empty XOR the target array is empty. The current implementation already does this if the target array is empty, but not if the original array is empty.

```csharp
	private static void PatchForArray(JsonArray original, JsonArray target, List<PatchOperation> patch, JsonPointer path)
	{
		if (original.Count == 0 ^ target.Count == 0)
		{
			patch.Add(PatchOperation.Replace(path, target));
		}
		else if (target.Count >= original.Count)
		{
			for (int i = 0; i < target.Count; i++)
			{
				if (i < original.Count)
				{
					CreatePatch(patch, original[i], target[i], path.Combine(i));
					continue;
				}
				patch.Add(PatchOperation.Add(path.Combine(i), target[i]));
			}
		}
		else
		{
			// Loop backwards because Remove operations cause array lengths to shrink.
			for (int i = original.Count - 1; i >= 0; i--)
			{
				if (i < target.Count)
				{
					CreatePatch(patch, original[i], target[i], path.Combine(i));
					continue;
				}
				patch.Add(PatchOperation.Remove(path.Combine(i)));
			}
		}
	}
```

If this sounds alright, I can submit a pull request.

_No response_

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

stephenmk:
### Description

This pull request adds an optimization to `PatchExtensions.PatchForArray`. When generating a patch in which an array from the original node contains 0 items and the corresponding target array contains N items, N > 0, then then the output patch should contain 1 `Replace` operation instead of N `Add` operations.

### Links

Resolves #964
Relates to #963 (also modifies this class method)

### Checks

### DIFF
--- a/tools/ApiDocsGenerator/release-notes/rn-json-patch.md
+++ b/tools/ApiDocsGenerator/release-notes/rn-json-patch.md
@@ -7,7 +7,7 @@ order: "09.09"
 # [4.1.0](https://github.com/gregsdennis/json-everything/pull/970) {#release-patch-4.1.0}
 
 - [#963](https://github.com/json-everything/json-everything/pull/963) - Runtime performance improvement for `PatchExtensions.CreatePatch` by [@stephenmk](https://github.com/stephenmk)
-- [#964](https://github.com/json-everything/json-everything/pull/970) - Patch size optimization for comparisons involving zero-length arrays by [@stephenmk](https://github.com/stephenmk)
+- [#964](https://github.com/json-everything/json-everything/issues/964) - Patch size optimization for comparisons involving zero-length arrays by [@stephenmk](https://github.com/stephenmk)
 
 # [4.0.0](https://github.com/gregsdennis/json-everything/pull/952) {#release-patch-4.0.0}
 


gregsdennis:
@stephenmk since you have two PRs open for Patch, please bump the patch version and add an entry in the release notes on one of them, then I'll merge them so that it publishes the new version.

PR #972, opened 2025-12-21
Trim leading periods from full type name (reapply)

gregsdennis:
@jochenkirstaetter I accidentally merged your #947.  It needed some updates.  Please update that here.


gregsdennis:
@jochenkirstaetter since you haven't updated this with the requested tests, I'm closing this.  Tests are needed both to ensure compliant behavior and to validate that the change is needed.

PR #1032, opened 2026-04-17
fix(playground): light mode cursor invisible due to missing color-scheme

### LINKED ISSUE #983: Light mode on playground uses white cursor over text entry controls
### Nuget Package

json-everything.net (playground)

### Package Version

n/a

### Operating System

None

None

The cursor doesn't respect the theme.

### What did you expect?

The cursor should respect the theme.

1. Go to the website.
2. Switch to light mode.
3. Move mouse over text entry field.

Cursor cannot be seen.

I spent about two hours working with AI trying to fix this.  I need someone who is a better FE developer than I am.

MasterOfFeelingFish:
The text cursor (I-beam) is white in light mode, making it invisible against the white background. Bootstrap 5.1 doesn't set `color-scheme` on the document element, so the browser defaults to dark-mode native controls regardless of the `data-bs-theme` attribute.

Three changes:
- Added `color-scheme: light dark` on `html` in `app.css` to declare both modes are supported
- Updated `ApplyTheme()` in `Header.razor` to set `document.documentElement.style.colorScheme` alongside the existing `data-bs-theme` toggle
- Set `style="color-scheme: dark"` on the `<html>` tag in `index.html` so the loading screen also renders correctly before Blazor initializes

Fixes #983

### DIFF
--- a/json-everything.net/Shared/Header.razor
+++ b/json-everything.net/Shared/Header.razor
@@ -100,7 +100,7 @@
 	private async Task ApplyTheme()
 	{
 		await JSRuntime.InvokeVoidAsync("eval", _isDarkMode 
-			? "document.documentElement.removeAttribute('data-bs-theme')" 
-			: "document.documentElement.setAttribute('data-bs-theme', 'light')");
+			? "document.documentElement.removeAttribute('data-bs-theme'); document.documentElement.style.colorScheme='dark'" 
+			: "document.documentElement.setAttribute('data-bs-theme', 'light'); document.documentElement.style.colorScheme='light'");
 	}
 }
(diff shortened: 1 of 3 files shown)


gregsdennis:
@MasterOfFeelingFish thanks for this, but I see no change when I run this locally.  The cursor is still white.

MasterOfFeelingFish:
Thanks for testing. It seems color-scheme alone doesn't override the native I-beam cursor color on your system. I'll need to dig deeper — this may require forcing the cursor via a Blazor/JS interop approach or a different CSS strategy. I'll follow up when I have a working fix.

Regarding the CLA: I'll need to review the agreement before signing. Will handle that separately.

gregsdennis:
My system is set to dark mode.  That may have something to do with it, but I've tried setting it to light mode, and I get the same result.

